PR #1, opened 2017-06-25
Remove stale advice from README.md

kaduk:
After commit 8909d2161299bafbdb0690cdf8b677c0ce7c2f34, gitaddrev
uses OpenSSL::Query to determine valid reviewers, so there is no
longer a hardcoded list to update.


richsalz:
merged.

PR #2, opened 2017-07-04
Add a little safety to merge scripts

paulidale:
Make the final question before push have a _yes_/_no_ answer.

The response is treated case insensitively and short forms are accepted (_y_ and _n_).


richsalz:
merged; thanks for the improvements.

PR #3, opened 2017-07-26
Modifications to pick-to-branch script.

paulidale:
Minor changes to the pick-to-branch script.
    
Change the exit in the middle to a yes/no question asking if the script should continue.
    
Change "git co" to "git checkout".  The former fails here for some reason.


richsalz:
merged, thanks.

PR #6, opened 2017-12-08
Add an option to not add yourself as a reviewer

mattcaswell:
Every now and then I need to merge a commit that I did not author or
review (someone else has done those things). This adds an option to addrev
to enable you to add the "Reviewed by" headers without also adding
yourself.

### DIFF
--- a/review-tools/addrev
+++ b/review-tools/addrev
@@ -9,6 +9,7 @@ my $list_reviewers = 0;
 my $help = 0;
 my $haveprnum = 0;
 my $trivial = 0;
+my $useself = 1;
 
 my $my_email;
 
@@ -23,6 +24,8 @@ foreach (@ARGV) {
         $args .= "--trivial ";
     } elsif (/^--verbose$/) {
         $args .= "--verbose ";
+    } elsif (/^--noself$/) {
+        $useself = 0;
     } elsif (/^--myemail=(.+)$/) {
         $my_email = $1;
     } elsif (/^--nopr$/) {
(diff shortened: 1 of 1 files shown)


levitte:
As far as I recall, we discussed exactly this case a few years ago, and we concluded that if you merge a certain PR, you should count yourself as a reviewer...  It does put some pressure on the one merging to have a look, at least a glimpse.  Do we need to reopen that discussion?  I can certainly see reasons to...

mattcaswell:
Generally speaking it is the case the mergers are also reviewers - and I think that is a good general "rule-of-thumb". However there are cases where this isn't is the right approach. Most often I encounter this when doing a security release. I have a number of commits to merge for different issues. They may have been authored and reviewed by other people. I don't feel a need as the the person doing a release to review all commits (as long as I know someone has).

richsalz:
What about using -s (signed off by) to indicate that you aren't a reviewer or author, but you pushed the code?  Without something like that, we have no way of knowing who actually did the merge.  As for me, I agree with @levitte's first comment, not his second. :)

mattcaswell:
> Without something like that, we have no way of knowing who actually did the merge. As for me, I agree with @levitte's first comment, not his second. :)

Git already tells you this without a need for additional headers. Try `git log --pretty=full` and it tells you who actually did the commit. I actually have this in my global git config:

format.pretty=fuller

So I get the "fuller" format by default. That shows me the same stuff as "full" (who did the commit) as well as author and commit dates which I find particularly useful.

richsalz:
Thanks; I withdraw my concern.

richsalz:
@mattcaswell are you going to merge this?

mattcaswell:
> @mattcaswell are you going to merge this?

Yes! Thanks for the reminder. Done.

PR #8, opened 2018-02-08
run-checker: enhance README

levitte:


levitte:
Merged. da7d7b372dc15d739df3ba7aff2c1a4292148515

I hope this is good enough to make the use of run-checker.sh seem easy enough...

PR #9, opened 2018-02-08
Fix run-checker-autohooks

levitte:
The README wasn't updated from the time this was an internal set of scripts.  Also, REPORT_FROM and REPORT_RECIPIENT were still holding internal values.  Clearing these things make these hooks ready for public consumption.


levitte:
Merged.  7723f7c116e2024e727648fe369133b0fc8f70e5 and 9d9fdc1727e3ff90d25dbccb4c530312b897536f

PR #10, opened 2018-02-13
Don't sanity check the FIPS version during a release

mattcaswell:
This is no longer included in the opensslv.h header file for 1.1.1, so
we should not sanity check it.


mattcaswell:
Pushed. Thanks.

richsalz:
We don't have a review policy for anything other than the main repo; in particular web and tools don't need it.  Still it's good form sometimes. :)

PR #11, opened 2018-02-13
tls1_3 is now on by default so we should instead test no-tls1_3

mattcaswell:


richsalz:
ping @mattcaswell and @levitte

mattcaswell:
Pushed. Thanks for the reminder.

richsalz:
Does @levitte just run the script out of the repo, or does he have to edit something else?  Hence the ping to him.

levitte:
I need to perform a `git pull --rebase`, 's all.

levitte:
Done

PR #13, opened 2018-05-21
Add --web flag for web repo

richsalz:
Adds support for the --web flag to push to that repo.
Should probably also have a --tools flag; maybe separate PR.


richsalz:
Thanks.  Coming up next --tools and perhaps getopt(1) use.

PR #20, opened 2018-07-11
CC default fix

paulidale:
The default for CC wasn't exporting the environment variable.  Likewise, the variable wasn't used.  Hence it didn't take effect.

This fixes the issue by exporting the environment variable in this case.


richsalz:
I don't think you have commit perms on the repo, so let me know and I'll do it.

paulidale:
I don't have the required permissions.

richsalz:
merged, thanks for the fix!

PR #22, opened 2018-07-14
run-checker: Add script to build gost.so

### LINKED ISSUE #21: Run run-checker with GOST test switched on
Now that openssl/openssl#6650 has been merged we have a test capable of testing GOST ciphersuites if a GOST engine in present. The test requires that the environment variable OPENSSL_GOST_ENGINE_SO is set to point at gost.so. We should set up run-checker so that this is configured.

levitte:
It gets built against a 1.1.0 build, on purpose to check that it links
and runs correctly against any libcrypto.so.1.1.

Fixes #21

### DIFF
--- a/run-checker/build-gost.sh
+++ b/run-checker/build-gost.sh
@@ -1,0 +1,39 @@
+#! /bin/bash
+#
+# Run in a directory for a gost engine build.
+# Two subdirectories will be created:
+#
+#    gost-engine	a checkout of https://github.com/gost-engine/engine.git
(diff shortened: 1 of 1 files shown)


levitte:
The crontab entries will be changed appropriately to run this script and to set OPENSSL_GOST_ENGINE_SO when running the run-checker script.

levitte:
Merged.

45ea2185ce1a2f39c6fb72abd076dbb191b0d016 run-checker: Add script to build gost.so

PR #23, opened 2018-07-16
build-gost: ensure gost.so is built with RUNPATH, not RPATH

levitte:
This allows us to point to whatever libcrypto we want.


levitte:
Arguably, this RUNPATH should be used by default, to reasonably ensure that whatever libcrypto the application uses is also the one used by gost.so.  Normally, that would be the system installation anyway, so no biggie...

levitte:
Merged.

1fb2c1d2155c25b18b00bc9cc6f1a9169d1b9460 build-gost: ensure gost.so is built with RUNPATH, not RPATH

PR #25, opened 2018-07-23
Don't test no-md5

mattcaswell:
no-md5 is not a valid option so we shouldn't be testing it


mattcaswell:
Note, due to a bug in Configure this invalid option was tolerated before. Now it isn't and causes run-checker to fail.

mattcaswell:
I merged this. @levitte, I think we need to update the run-checker box to use this?

PR #27, opened 2018-08-16
The "tls13downgrade" Configure option no longer exists

mattcaswell:
So, we need to remove checking of it from run-checker.


mattcaswell:
Pushed. Thanks.

PR #28, opened 2018-09-04
gitaddrev: don't lowercase the reviewers

levitte:
If we lowercase them, the occasional registered mixed case ID will not
be recognised.


levitte:
Merged.

1f45f077ae95e501865207cc9aba81a7c3791756 gitaddrev: don't lowercase the reviewers

PR #29, opened 2018-09-10
release-tools: use 'make tar' instead of 'make dist'

levitte:
'make dist' will pre-configure the OpenSSL source for distribution
with a simple generic configuration, which is nice...  as long as
you're on Unix.

Unfortunately, the resulting Makefile will be picked up by NMAKE (on
Windows) and MMS / MMK (on VMS) and will spew out errors because those
are very different environments.

Therefore, it's better not to pre-configure, and just create an
archive from the source proper.

(note that pre-configuration would still work on the 1.0.2 series,
since the Windows and VMS builds are entirely different there, but
it's no big loss to force everyone to configure when using 1.0.2 as
well)

### DIFF
--- a/release-tools/README.md
+++ b/release-tools/README.md
@@ -276,7 +276,7 @@ or if you want to use the openssl-team key:
 
 Make the release tarball. You do this with:
 
-        make dist
+        make tar
 
 Create .sha1, .sha256 and .asc files manually. You can use:
 
--- a/release-tools/mkrelease.pl
+++ b/release-tools/mkrelease.pl
@@ -257,7 +257,7 @@ if ( !$no_release ) {
     my $gpgann = $ENV{"OPENSSL_GPG_ANNOUNCE"}
       // "$gpg --use-agent -sta --clearsign";
     my $tarfile = "openssl-${expected_version}.tar.gz";
-    system("(./config; make $TAR dist) >../$tarfile.log 2>&1");
+    system("(./config; make $TAR tar) >../$tarfile.log 2>&1");
     die "Error generating release!" if $?;
     die "Can't find tarball!!" unless -f "../$tarfile";
 


levitte:
@openssl/omc, I think we did talk about this, but can't quite recall the concensus...

mattcaswell:
Could this be considered a breaking change for 1.0.2 and 1.1.0?

levitte:
*All* our build documentation essentially says "configure first, make second", so it will possibly surprised those who "discovered" our habit of making a pre-configured distribution.  We might have a few complaints, but I think they can easily be dismissed with "please configure first".  Most people do that anyway, since they want to take advantage of their platforms.

So, errr, naaah

levitte:
I was considering if I should add an exception for anything older than 1.1.0.  If you prefer that, I can still do so.

mattcaswell:
I think this is ok for 1.1.1. I'm less sure about 1.1.0/1.0.2. What do others think?

levitte:
I would say that for 1.1.0, we should switch to this, because the pre-configured Makefile is a disaster on non-Unix platforms.

richsalz:
I would do just master as it's the easiest and the other releases aren't going to be supported for very much longer.

levitte:
With master, you mean 1.1.1?

richsalz:
yes.  or just change the calling script.

levitte:
OK... So are we getting to some kind of decision here? I'm siding with @vdukhovni on this...

t-j-h:
I agree with @vdukhovni

levitte:
So I'm hoping that someone will approve this change, then.  It would be disappointing if it had to wait 'til after the 1.1.1 release, just sayin'...

levitte:
Thank you :-)

Merged.

eb4d1f2af762c9674d1461ddbe652f18bcd21415 release-tools: use 'make tar' instead of 'make dist'

PR #35, opened 2018-11-18
Tool changes for Mac OS/X.

paulidale:
Perl ends up with a different path when installed using _brew_.  The system perl
seems to have broken _cpan_ somehow.

Some additional packages need to be installed.

PATH also needs to be set so it can find the _plackup_ executable.

### DIFF
--- a/OpenSSL-Query/README.md
+++ b/OpenSSL-Query/README.md
@@ -23,6 +23,8 @@ OpenSSL::Query requires these extra modules to run:
 - Class::Method::Modifiers	(debian package libclass-method-modifiers-perl)
 - Moo				(debian package libmoo-perl)
 - URI::Encode			(debian package liburi-encode-perl)
+- LWP::UserAgent
+- LWP::Protocol::https
 
 Any other module OpenSSL::Query depends on should be part of core
 perl.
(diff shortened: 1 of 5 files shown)


OpenSSL-Query/README.md
```diff
@@ -23,6 +23,8 @@ OpenSSL::Query requires these extra modules to run:
 - Class::Method::Modifiers	(debian package libclass-method-modifiers-perl)
 - Moo				(debian package libmoo-perl)
 - URI::Encode			(debian package liburi-encode-perl)
+- LWP::UserAgent
+- LWP::Protocol::https
```
levitte: Please change the "requires" section in `Makefile.PL` as well

paulidale: Added.
I'm not au fait with Perl.

paulidale:
PS: you shouldn't be working on a Sunday :(
Especially this early.

OpenSSL-Query/README.md
paulidale: I've updated the required section (I think).

levitte:
Merged.

1de20a3a49e1b4f6e7104b84b8fc5594b144712a add dependencies to Makefile.PL
cd0711f60154d5da0f9974349b897a2615934554 Tool changes for Mac OS/X.

PR #37, opened 2019-03-14
run-checker: Check enabling and disabling trace

levitte:


InfoHunter:
What is the default behaviour of trace?

mspncp:
It’s currently changing from enabled to disabled, see #8474.

levitte:
It's precisely because we're currently in a state of fluctuation that I made it both.

(side note: when linking between pull requests in a different repo, you have to be a little more verbose: openssl/openssl#8474)

mspncp:
Thanks, I know. But I was doing it on my mobile (once again), so I was too lazy :P

mspncp:
> It's precisely because we're currently in a state of fluctuation that I made it both.

Since openssl/openssl#8474 has been merged meanwhile, I'd prefer having only the non-default now.

levitte:
Sure

levitte:
Merged.

31b2767973d7a072a36c0eaf68861f9a1a3fe7b2 run-checker: Check enabling and disabling trace

PR #42, opened 2019-08-28
Fix enable-fuzz-afl builds in run-checker

mattcaswell:
We now require no-module for successful enable-fuzz-afl builds


mattcaswell:
See also https://github.com/openssl/openssl/pull/9721

mattcaswell:
Pushed. Thanks.

PR #40, opened 2019-08-13
Add new ghlink tool

mspncp:
## The Story

Very often, when posting a reply to an issue or a pull request, It happens that I need to look up some stuff in my local repository:

```make
~/src/openssl$ git grep -nw stuff
.gitignore:69:# Fuzz stuff.
...
config:344:# Do the Apollo stuff first. Here, we just simply assume
...
crypto/aes/asm/vpaes-armv8.pl:94://  Decryption stuff
...
crypto/asn1/tasn_enc.c:316:        /* And the stuff itself */
...
crypto/bio/b_dump.c:11: * Stolen from tjh's ssl/ssl_trc.c stuff.
...
```

Grepping the stuff is the easy part, but after that it gets tedious, because I have to fire up my browser and navigate manually through the GitHub repository in order to retrieve the same locations once more. In these moments I dream of a tool that could convert those grep hits into URLs, which would take me off to GitHub via  `<ctrl>-<left-mouse-click>` :rocket:

```make
~/src/openssl$ git grep -nw stuff
https://github.com/openssl/openssl/blob/master/.gitignore#L69 # Fuzz stuff.
...
https://github.com/openssl/openssl/blob/master/config#L344 # Do the Apollo stuff first. Here, we just simply assume
...
https://github.com/openssl/openssl/blob/master/crypto/aes/asm/vpaes-armv8.pl#L94 //  Decryption stuff
...
https://github.com/openssl/openssl/blob/master/crypto/asn1/tasn_enc.c#L316         /* And the stuff itself */
...
https://github.com/openssl/openssl/blob/master/crypto/bio/b_dump.c#L11  * Stolen from tjh's ssl/ssl_trc.c stuff.
...
```

### Dreaming further...

...I whish I could also post these links as markdown links...

> [.gitignore:69:](https://github.com/openssl/openssl/blob/master/.gitignore#L69) # Fuzz stuff.
> [CHANGES:11033:](https://github.com/openssl/openssl/blob/master/CHANGES#L11033)      the new code. Add documentation for this stuff.
> [config:344:](https://github.com/openssl/openssl/blob/master/config#L344) # Do the Apollo stuff first. Here, we just simply assume
> [crypto/aes/asm/vpaes-armv8.pl:94:](https://github.com/openssl/openssl/blob/master/crypto/aes/asm/vpaes-armv8.pl#L94) //  Decryption stuff
[...]

PR #39, opened 2019-08-01
QueryApp: enhance documentation of local installation via local::lib

mspncp:
@levitte this is a little "thank you" for you for fixing my addrev problems and pointing me to local::lib. Hope you like it.

PR #33, opened 2018-10-24
cherry-checker: add a 'fixes' column

mspncp:
Scans the commit messages for 'Fixes' annotations and displays them in an additional column.

Example output:

```
$ cherry-checker  --all --remote

These cherries are hanging on the git-tree:

  <-  origin/master
  ->  origin/OpenSSL_1_1_1-stable
  ==  both

 prnum | fixes | br |   commit   |   subject
 ----- | ----- | -- | ---------- | -------------------------------------------
 #7467 |       | == | 28361a0b82 | RAND: ensure INT32_MAX is defined 
 #7467 |       | == | f81b043ad8 | RAND: ensure INT32_MAX is defined 
 #7352 |       | <- | 97b0b713fb | RSA security bits calculation 
 #7455 | #7449 | == | ece482ff3a | RAND_add(): fix heap corruption in error path 
 #7455 | #7449 | == | 5b4cb385c1 | RAND_add(): fix heap corruption in error path 
 #7451 |       | -> | 132fd512ad | build file templates: have targets for all shared library names 
 #7451 |       | <- | d8cac50b02 | build file templates: have targets for all shared library names 
 #7434 | #7433 | == | 8c6371f9f7 | Don't complain and fail about unknown TLSv1.3 PSK identities in s_serv... 
 #7434 | #7433 | == | 2d015189b9 | Don't complain and fail about unknown TLSv1.3 PSK identities in s_serv... 
 #7375 | #6934 | == | d1bfd8076e | Buffer a ClientHello with a cookie received via DTLSv1_listen 
 #7375 |       | == | 585e691948 | Use the read and write buffers in DTLSv1_listen() 
 #7375 | #6934 | == | 079ef6bd53 | Buffer a ClientHello with a cookie received via DTLSv1_listen 
 #7375 |       | == | 2fc4c77c3f | Use the read and write buffers in DTLSv1_listen() 
 #7431 |       | == | 6c529877cd | Test DTLS cookie generation and verification 
 #7431 | #7428 | == | a6a83827a0 | Fix a DTLS memory leak 
 #7431 |       | == | edcd29efd3 | Test DTLS cookie generation and verification 
 #7431 | #7428 | == | 01666a8c1d | Fix a DTLS memory leak 
 #7294 |       | <- | 9986bfefa4 | sha/asm/keccak1600-armv8.pl: halve the size of hw-assisted subroutine. 
 #7400 |       | == | a66c361a77 | Configurations/15-android.conf: add support for "standalone toolchain"... 
[...]

PR #43, opened 2019-10-03
cherry-checker: fix: don't choke on semicolons in the commit message title

mspncp:
The semicolon is used as separator for the git output, so make sure
to stop the line.split() after the third separator, otherwise the
subject line might get split further.

(see openssl/openssl@bd01733fdd Fix comment; unchecked->checked)

### DIFF
--- a/review-tools/cherry-checker
+++ b/review-tools/cherry-checker
@@ -88,7 +88,7 @@ def pick_cherries(left, right, all = False):
 
     for line in subprocess.check_output(git_command).decode().splitlines():
 
-        timestamp, branch, commit, subject = line.split(";")
+        timestamp, branch, commit, subject = line.split(";", maxsplit=3)
 
         if branch == '=' and not all:
             continue


mspncp:
Thank you Pauli. Just as a reminder: I don't have commit access, would you be so kind and do the merge please?

paulidale:
I think it technically requires a second OMC review before merging.

mspncp:
I'm not sure whether the same formal rules apply. Most often it was only @levitte who approved (see #40 #39 #33).

paulidale:
Okay, I was being cautious.  Merged now.

mspncp:
Well, it seems like I persuaded you to bypass the official rules. Sorry for that, it was not on purpose.

t-j-h:
Approved (for avoid doubt on the context).

paulidale:
Yeah, you did.  I should have checked further.  It's my fault ultimately.

Thanks @t-j-h for the confirmation.

mattcaswell:
The git hooks should have caught it and prevented the push, but I guess they've not been adjusted to understand the difference between the main source repo and web/tools.

paulidale:
I don't think they restrict pushes to the other repos at all.
Apart from OMC doing it.

mattcaswell:
Hmmm. On checking, the hook to verify the "Reviewed-by" headers only runs on the openssl and openssl-book repos.

PR #47, opened 2019-10-25
clacheck: rename 'need-cla' label to 'hold: cla required'

mspncp:
It think this is all what needs to be changed for clacheck, but I can't test it.

### DIFF
--- a/clacheck/clacheck.py
+++ b/clacheck/clacheck.py
@@ -20,6 +20,8 @@ SUCCESS = 'success'
 FAILURE = 'failure'
 CLAFILE = "/var/cache/openssl/checkouts/omc/cladb.txt" #<EDIT>
 
+CLA_LABEL='hold: cla required'
+
 null_actions = (
         'assigned', 'unassigned', 'labeled', 'unlabeled', 'closed',
         'review_requested', 'review_request_removed',
@@ -55,14 +57,14 @@ def update_status(pr, state, description):
     conn.getresponse().read()
     host,url = url_split(pr['issue_url'])
     if state == SUCCESS:
-        url = url + '/labels/need-cla'
+        url = url + '/labels/'+ urllib.quote(CLA_LABEL)
         print 'Delete', url
         conn.request('DELETE', url, None, headers)
     elif state == FAILURE:
         url = url + '/labels'
         print 'Add need-cla', url
         conn.set_debuglevel(99)
-        conn.request('POST', url, '[ "need-cla" ]', headers)
+        conn.request('POST', url, '[ "{}" ]'.format(CLA_LABEL), headers)
     reply = conn.getresponse().read()
     print "--\n", reply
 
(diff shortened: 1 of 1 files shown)


mspncp:
cc @levitte

clacheck/clacheck.py
```diff
@@ -20,6 +20,8 @@
 FAILURE = 'failure'
 CLAFILE = "/var/cache/openssl/checkouts/omc/cladb.txt" #<EDIT>
 
+CLA_LABEL='hold: cla required'
```
levitte: Nit: perhaps do like in the rest of the file, have a little bit of spacing around the assignment operator?

clacheck/clacheck.py
```diff
@@ -55,14 +57,14 @@ def update_status(pr, state, description):
     conn.getresponse().read()
     host,url = url_split(pr['issue_url'])
     if state == SUCCESS:
-        url = url + '/labels/need-cla'
+        url = url + '/labels/'+ urllib.quote(CLA_LABEL)
```
levitte: Same here, space before the `+` too

levitte:
Other than those two nits, that looks right to me.

clacheck/clacheck.py
mspncp: Done in ada6caa4888b12ae3aa10b97e8343a3f832df291.

clacheck/clacheck.py
mspncp: Ups, didn't read far enough. Amended and force-pushed my fixup, see ac4e41a1386419e044448b019e83e575133523df.

mspncp:
Note: When this has been merged, you might want to change the label yourself on [openssl/labels](https://github.com/openssl/openssl/labels). Instead of moving all tickets (including the closed ones!), you can just delete the [hold: cla required]  label and rename [need-cla] to [hold: cla required], both labels already have the some color.

mspncp:
BTW: my implicit approval does not count in this repo. So it needs a second OMC review.

t-j-h:
Nothing quite like reviewing while seated on a plane waiting for take off :-)

mspncp:
Since the same policy as for openssl/openssl applies to also to openssl/tools and openssl/web analogously, we should add a decent subset of the new labels to those as well.

levitte:
Merged.

73b0391902c67f9ebfcd5fa96cc10734d9631da8 clacheck: rename 'need-cla' label to 'hold: cla required'

levitte:
Side note: I fixed all deployment keys so they were added by @openssl-machine instead of me.  Hence the push that is shown.

mspncp:
@levitte will you rename the need-cla label, or shall I do it?

mspncp:
> @levitte will you rename the need-cla label, or shall I do it?

I'll just do it and hope for no race condition... ;-)

mspncp:
Done :-)

PR #50, opened 2020-01-05
Apply OTC changes to gitaddrev

romen:
After the changes in https://github.com/openssl/web/pull/146
`gitaddrev` (on which `addrev` depends) required an update to count OTC
approvals rather than OMC approvals.

### DIFF
--- a/review-tools/gitaddrev
+++ b/review-tools/gitaddrev
@@ -29,7 +29,7 @@ my @reviewers;
 my @nocla_reviewers;
 my @unknown_reviewers;
 my $skip_reviewer;
-my $omccount = 0;
+my $otccount = 0;
 sub try_add_reviewer {
     my $id = shift;
     my $rc = undef;
@@ -39,7 +39,7 @@ sub try_add_reviewer {
 	my $cla = $query->has_cla(lc $rev);
 	if ($cla) {
 	    unless (grep {$_ eq $rev} @reviewers) {
-		$omccount++ if $query->is_member_of($id2, 'omc');
+		$otccount++ if $query->is_member_of($id2, 'otc');
 		push @reviewers, $rev;
 	    }
 	    $rc = $rev;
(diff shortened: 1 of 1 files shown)


romen:
My Perl-foo is as usual very close to zero, so please have an in-depth look to see if what I did makes sense, as I don't fully grasp all the details of how things are checked!

levitte:
Side note: -fu, not -foo.

https://en.m.wiktionary.org/wiki/-fu

mattcaswell:
How does this impact the review headers for other repos? So, with addrev you can specify --web or --tools to specify that you are applying headers for a different repo than the main code one.

For that matter what are the review rules for other repos. For the code repo it is clear. But what about web and tools? It kind of makes sense at least for tools to be OTC owned?

romen:
> How does this impact the review headers for other repos? So, with addrev you can specify --web or --tools to specify that you are applying headers for a different repo than the main code one.
> 
> For that matter what are the review rules for other repos. For the code repo it is clear. But what about web and tools? It kind of makes sense at least for tools to be OTC owned?

I believe OMC should clearly state this officially somewhere to clear the ambiguity: personally I would say that `web` should not fall under OTC control as I believe the fundamental role of the website goes beyond the area of interest of the OTC and more under the direct functions of the OMC.

`tools` seems to include mostly stuff that appears pretty much of OTC interest, but maybe at least  `tools/QueryApp` should be exempt from OTC control?

That's of course just my personal opinion, and I only use a tiny fraction of the tools in `tools`!


That being said, until your comment I did not even thing that `addrev` was used also outside of reviews for the `openssl/openssl` repo, so maybe @levitte could suggest something better for handling repos under OTC or under OMC control?

Anyway, until the OMC states otherwise, this PR must be merged by an OMC member.

paulidale:
@mattcaswell didn't we discuss some of the ownership at the F2F?  I've a recollection that at least one of the repos looked like it should be jointly owned (i.e. needed splitting).

mattcaswell:
Pushed.

PR #58, opened 2020-02-11
Improve tools::OpenSSL-Query/README.md

DDvO:
adds missing dependency info to the Testing section
allows simple copy&paste for the series of shell commands

### DIFF
--- a/OpenSSL-Query/README.md
+++ b/OpenSSL-Query/README.md
@@ -32,9 +32,8 @@ perl.
 Installation
 ------------
 
-    $ perl Makefile.PL
-    $ make
-    $ make install
+    perl Makefile.PL
+    make && make install
 
 Local installation
 ------------
@@ -43,7 +42,7 @@ For a local installation, you might want to consider using local::lib
 (debian package liblocal-lib-perl).  In that case, running Makefile.PL
 is slightly different:
 
-    $ perl -Mlocal::lib Makefile.PL
+    perl -Mlocal::lib Makefile.PL
 
 Other than that, follow the instructions in "Installation" above.
 
@@ -58,12 +57,12 @@ Testing
 
 Testing is done like this:
 
-    $ make test
+    make test
 
 However, it requires that a temporary query service is started as
 well.  This is part of QueryApp, and is started like this:
 
-    $ here=`pwd`
-    $ cd ../QueryApp	# Or wherever you have it checked out
-    $ PERSONDB=./t/query_data/pdb.yaml CLADB=./t/query_data/cdb.txt \
-      PERL5LIB=./lib:$here/lib plackup bin/query.psgi
+    here=`pwd`        # the directory OpenSSL-Query/
+    cd ../QueryApp	# Or wherever you have it checked out
+    PERSONDB=./t/query_data/pdb.yaml CLADB=./t/query_data/cdb.txt \
+    PERL5LIB=./lib:$here/lib plackup bin/query.psgi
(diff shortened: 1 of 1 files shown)


OpenSSL-Query/README.md
```diff
@@ -58,12 +57,14 @@ Testing
...
+plackup (debian packages libdancer2-perl and libplack-perl)
+and that a temporary query service is started as well.
+This is part of QueryApp, and is started like this:
 
-    $ here=`pwd`
-    $ cd ../QueryApp	# Or wherever you have it checked out
-    $ PERSONDB=./t/query_data/pdb.yaml CLADB=./t/query_data/cdb.txt \
-      PERL5LIB=./lib:$here/lib plackup bin/query.psgi
+    here=`pwd`        # the directory OpenSSL-Query/
+    cd ../QueryApp	# Or wherever you have it checked out
```
levitte: Nit: align the comments, and most of all, I prefer to see a little more distance between the command and the comment, for readability

DDvO: Ah, there was a nasty TAB character in between - removed.
I've also put more space as you suggested.

DDvO:
Since meanwhile more than a day has passed that this PR has receive two OTC approvals and has been marked 'ready' I've been trying to push this to master at openssl-git@git.openssl.org:tools.git

Yet I get write access denied:
```
Pushing to openssl-git@git.openssl.org:tools.git
FATAL: W any tools dev@ddvo.net DENIED by fallthru
(or you mis-spelled the reponame)
fatal: Could not read from remote repository.
```

Is maybe also here an inconsistency with my alternative email address,
or are only OTC members allowed to push to this repo?

romen:
This repo still belongs to OMC at the moment IIRC

levitte:
I've raised the ownership question with the rest of the OMC.  Meanwhile, I'm merging this now, for the sake of expedience.

levitte:
Merged

d250d42df4c99fd4ae57c75de514dcaf9ff96360 allow simple copy&paste for the series of shell commands in OpenSSL-Query/README.md
c217de2c9b91c88f64ce6940b6c3ef5ef1a0bdae add missing dependency info to the Testing section of OpenSSL-Query/README.md

levitte:
Argh...  I forgot to edit the messages to correct the merge URLs :-/

mattcaswell:
> I forgot to edit the messages 

You do know that addrev can write the correct URLs in the first place right? i.e. via the "--tools" option

DDvO:
> Argh... I forgot to edit the messages to correct the merge URLs :-/

Thanks @levitte for taking over the merge.
I don't see anything wrong with it :)

levitte:
> You do know that addrev can write the correct URLs in the first place right? i.e. via the "--tools" option

I actually hadn't discovered that...  and now I see `--web` too...  it's really not the best of interfaces, I would rather have something that looks at the remote or something like that.
I've been thinking that `addrev` needs a rewrite anyway, it makes too much duplicate work (*every* separate commit that's to be rewritten generates that *exact* same calls to the people DB, that's a huge waste of time)....  it won't happen immediately, though.

levitte:
> I don't see anything wrong with it :)

```
    (Merged from https://github.com/openssl/openssl/pull/58)
```

Have a hard look.  I totally expect an "Ah-ha!!!" from you :wink:

DDvO:
Ah, yes, @levitte, I must have had tomatoes on my eyes :wink: 
(which is the literal translation of a German saying with obvious meaning)

DDvO:
BTW, with the enhances version of `ghmerge` I'm currently setting up in #59,
the merge could have been done very conveniently in a little error-prone manner like this:
```
ghmerge --tools --nomerge 58 paulidale levitte
```

levitte:
...  ok.  I promise I'll have a closer look, but do expect scepticism until I warm up, 'k?

DDvO:
Meanwhile, the above example use of `ghmerge` has simplified to
```
ghmerge --tools 58 paulidale levitte
```

PR #59, opened 2020-02-11
Improve addrev and ghmerge tools and their README

DDvO:
Various tweaks to make `ghmerge` more usable:
    * make gcc (rather than clang-3.6) the default for rebuilding
    * take the (SSH version of the) repo URL from the database
    * print some more (and slightly improve) info what is going on
    * show the log of the commits be pushed (after `addrev` is called)

Extend `review-tools/README`, making clear that `ghmerge` calls `addrev`

BTW, @mattcaswell, thanks for the Committer Instructions I had received by email.
I suggest adapting also these such that it becomes clear that  `ghmerge` uses `addrev`.
For instance:

```
Example usage adding PR and reviewer info to the last two commits:
$ addrev -2 --prnum=3711 matt @levitte master..

Example usage adding PR and reviewer info to the last commit and performing a non-merge push:
$ ghmerge --nomerge 3711 matt @levitte 
```

### DIFF
--- a/review-tools/README
+++ b/review-tools/README
@@ -47,10 +47,10 @@ The transfer may take many seconds, in particular with the '--list' option.
 
 Examples:
 
-    addrev steve
-    addrev -2 steve
-    addrev -2 steve @richsalz
-    addrev -2 --reviewer=steve --reviewer=rsalz@openssl.org
+    addrev --prnum=1234 steve
+    addrev 1234 -2 steve
+    addrev 1234 -2 steve @richsalz
+    addrev 1234 -2 --reviewer=steve --reviewer=rsalz@openssl.org
 
 gitlabutil
 ----------
--- a/review-tools/README
+++ b/review-tools/README
@@ -117,21 +117,18 @@ The environment variables 'https_proxy' and 'no_proxy' can be used.
 
 Example usage patterns:
         ghmerge <prnum> <reviewers>...
-        ghmerge --tools --nomerge <prnum> <reviewer>...
+        ghmerge --tools --squash <prnum> <reviewer>...
 
 Available options are:
 
---nomerge         Use 'git rebase -i --autosquash'.
-                  This is the default.
---autosquash      An alias to '--nomerge'.
---merge           Use 'git merge --no-commit --squash'.
---squash          An alias to '--merge'.
+--squash          Use non-interactive 'git merge --ff-only --squash'.
+                  The default is interactive 'git rebase -i --autosquash'.
 --nobuild         Do not use 'opensslbuild'.
                   Else it is invoked with $CC defaulting to "ccache gcc".
 --tools           Select the 'tools' repository rather than 'openssl'.
-                  This implies '--merge' and '--nobuild'.
+                  This implies '--nobuild'.
 --web             Select the 'web' repository rather than 'openssl'.
-                  This implies '--merge' and '--nobuild'.
+                  This implies '--nobuild'.
 --trivial         Pass '--trivial' to the invocation of 'addrev'.
 
 The <prnum> is the GitHub PR number.  The rest of the args are the names
--- a/review-tools/README
+++ b/review-tools/README
@@ -121,17 +121,19 @@ Example usage patterns:
         ghmerge <prnum> <reviewers>...
         ghmerge --tools --squash <prnum> <reviewer>...
 
+The default commit post-processing operaton is 'git rebase -i --autosquash'.
+
 Available options are:
 
---squash          Use non-interactive 'git merge --ff-only --squash'.
-                  The default is interactive 'git rebase -i --autosquash'.
---nobuild         Do not use 'opensslbuild'.
-                  Else it is invoked with $CC defaulting to "ccache gcc".
---tools           Select the 'tools' repository rather than 'openssl'.
-                  This implies '--nobuild'.
---web             Select the 'web' repository rather than 'openssl'.
-                  This implies '--nobuild'.
---trivial         Pass '--trivial' to the invocation of 'addrev'.
+--noautosquash  Use default interactive post-processing but without '--autosquash'.
+--squash        Use non-interactive post-processing 'git merge --ff-only --squash'.
+--nobuild       Do not use 'opensslbuild'.
+                Else it is invoked with $CC defaulting to "ccache gcc".
+--tools         Select the 'tools' repository rather than 'openssl'.
+                This implies '--nobuild'.
+--web           Select the 'web' repository rather than 'openssl'.
+                This implies '--nobuild'.
+--trivial       Pass '--trivial' to the invocation of 'addrev'.
 
 The <prnum> is the GitHub PR number.  The rest of the args are the names
 of the reviewers. All this info will be passed to addrev; see above.
--- a/review-tools/README
+++ b/review-tools/README
@@ -101,16 +101,37 @@ Checkout branch for query 145:
 ghmerge
 -------
 
-ghmerge merges (reviewed and approved!) GitHub pull requests
+ghmerge calls addrev and pushes (reviewed and approved!) GitHub pull requests.
+It includes several safety precautions and questions such as showing the diff,
+showing the resulting commit messages, and (by default) rebuilding everything.
 
-It works on the current branch, which should be master or one of the stable
-releases.
+It works on the current branch, which should be 'master' or one of the stable
+releases. For the branch to be merged a suitable remote must be available.
 
-Usage:
-        ghmerge ### reviewer...
+The tool accesses external databases via HTTPS (e.g., on api.github.com).
+The environment variables 'https_proxy' and 'no_proxy' can be used.
 
-The ### is the GitHub MR number.  The rest of the args are the names of
-the reviewers (passed to addrev; see above).
+Example usage patterns:
+        ghmerge <prnum> <reviewers>...
+        ghmerge --tools --nomerge <prnum> <reviewer>...
+
+Available options are:
+
+--nomerge         Use 'git rebase -i --autosquash'.
+                  This is the default.
+--autosquash      An alias to '--nomerge'.
+--merge           Use 'git merge --no-commit --squash'.
+--squash          An alias to '--merge'.
+--nobuild         Do not use 'opensslbuild'.
+                  Else it is invoked with $CC defaulting to "ccache gcc".
+--tools           Select the 'tools' repository rather than 'openssl'.
+                  This implies '--merge' and '--nobuild'.
+--web             Select the 'web' repository rather than 'openssl'.
+                  This implies '--merge' and '--nobuild'.
+--trivial         Pass '--trivial' to the invocation of 'addrev'.
+
+The <prnum> is the GitHub PR number.  The rest of the args are the names
+of the reviewers. All this info will be passed to addrev; see above.
 
 
 pick-to-branch
--- a/review-tools/README
+++ b/review-tools/README
@@ -113,7 +113,7 @@ It includes several safety precautions and questions such as showing the diff,
 showing the resulting commit messages, and (by default) rebuilding everything.
 
 It works on the current branch, which should be 'master' or one of the stable
-releases. For pushing the remote 'origin' is used by default.
+releases. The default remote is the first one matching 'git.openssl.org.*(push)'.
 So typically before calling 'ghmerge' one would have done the following:
 
 	git remote -v
@@ -138,7 +138,7 @@ Available options are:
 --nobuild       Do not use 'opensslbuild'.
                 Else it is invoked with $CC defaulting to "ccache gcc".
 --remote        Select the git remote of the branch to pull from and push to.
-                Default is 'origin'.
+                Default is the first remote matching 'git.openssl.org.*(push)'.
 --tools         Select the 'tools' repository rather than 'openssl'.
                 This implies '--nobuild'.
 --web           Select the 'web' repository rather than 'openssl'.
--- a/review-tools/README
+++ b/review-tools/README
@@ -11,7 +11,10 @@ the QueryApp libraries as well.
 Environment
 ===========
 
-Some of the scripts use the information REST API on https://api.openssl.org.
+Some of the scripts use the information REST API on https://api.openssl.org
+or on https://api.github.com.
+The environment variables 'https_proxy' and 'no_proxy' can be used.
+
 If you have direct access to the databases and want to use that instead, set
 the environment variable OMC to the directory where they are located.
 
@@ -38,8 +41,7 @@ IDs prefixed with a @, or known email addresses if given with --reviewer.
 
 Run 'addrev --list' to ge a list of known reviewer names.
 
-The tool accesses databases on www.openssl.org via HTTPS.
-The environment variables 'https_proxy' and 'no_proxy' can be used.
+The tool accesses databases on api.openssl.org via HTTPS.
 The transfer may take many seconds, in particular with the '--list' option.
 
 Examples:
@@ -112,7 +114,7 @@ showing the resulting commit messages, and (by default) rebuilding everything.
 It works on the current branch, which should be 'master' or one of the stable
 releases. For the branch to be merged a suitable remote must be available.
 
-The tool accesses external databases via HTTPS (e.g., on api.github.com).
+The tool accesses external databases on api.github.com via HTTPS.
 The environment variables 'https_proxy' and 'no_proxy' can be used.
 
 Example usage patterns:
--- a/review-tools/README
+++ b/review-tools/README
@@ -12,7 +12,7 @@ Environment
 ===========
 
 Some of the scripts use the information REST API on https://api.openssl.org
-or on https://api.github.com.
+while ghmerge also uses https://api.github.com.
 The environment variables 'https_proxy' and 'no_proxy' can be used.
 
 If you have direct access to the databases and want to use that instead, set
@@ -41,7 +41,8 @@ IDs prefixed with a @, or known email addresses if given with --reviewer.
 
 Run 'addrev --list' to ge a list of known reviewer names.
 
-The tool accesses databases on api.openssl.org via HTTPS.
+The tool accesses databases on api.openssl.org.
+The environment variables 'https_proxy' and 'no_proxy' can be used.
 The transfer may take many seconds, in particular with the '--list' option.
 
 Examples:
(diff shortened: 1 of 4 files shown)


review-tools/ghmerge
```diff
@@ -82,14 +88,19 @@ function cleanup {
...
 
-echo -n Press return to merge to $REL: ; read foo
 addrev $TRIVIAL --prnum=$PRNUM $TEAM ${REL}..
+echo Log since $REL
+git log $REL..
+
+echo -n Press return to merge to $REL: ; read foo
 git checkout $REL
 if [ "$MERGE" == "yes" ] ; then
     git merge --no-commit --squash $WORK
```
paulidale: Not directly related this, but I'd like to have something more akin to `git rebase -i --autosquash` here.  Being able the tune the squashing is the cause of the majority of times I don't use `ghmerge`.

DDvO:
I've just added also the following info on `(git)addrev` in `review-tools/README`:

```
The tool accesses databases on www.openssl.org via HTTPS.
The environment variable 'https_proxy' can be used to connect via a proxy.
The transfer may take many seconds, in particular with the '--list' option.
```

review-tools/ghmerge
```diff
@@ -100,12 +111,12 @@ else
...
 if [ "$BUILD" == "yes" ] ; then
-    echo Rebuilding
-    ( opensslbuild 2>&1 ) | tail -3
+    echo Rebuilding...
+    ( CC="$CC" opensslbuild 2>&1 ) | tail -3
 fi
 
 while true ; do
-    echo -n "Enter YES to push or NO to abort: "
+    echo -n "Enter YES/yes to push or NO/no to abort: "
```
paulidale: I'm not sure this change is all that useful.

BTW: `pick-to-branch` also has the same yes/no question.

review-tools/ghmerge
DDvO: Good point. 
Since you already approved this PR, I suggest you do this in a separate PR.
If you prefer, I can `s/git merge --no-commit --squash/git rebase -i --autosquash/` here.

review-tools/ghmerge
DDvO: I just wanted to make clear that the simpler-to-type lower-case variants can also be used.
How if I instead `s/YES/yes/` and `s/NO/no/` in this tool and in `review-tools/pick-to-branch`?

review-tools/ghmerge
DDvO: Now I see that `-i --autosquash` should better be added to the existing `git rebase` variant.
I've done this and updated and slightly extended the README accordingly.

review-tools/ghmerge
DDvO: I've just implemented the latter.

levitte:
While you're tweaking, would it be possible to have `$MERGE` default to "no"?  The mere fact that this script defaults to squashing all commits into one has had me avoid it like the plague.

review-tools/ghmerge
paulidale: That's a better suggestion.  I end up typing just _y_ usually.
The intent of the question was to stop accidental continuation, but I do answer _no_ from time to time.

review-tools/ghmerge
paulidale: Thanks, here is best since there are a number of other changes underway.
I'll clear my review.

paulidale:
A strong "yes" to defaulting `$MERGE` to no.

review-tools/ghmerge
```diff
@@ -59,9 +60,14 @@ set -- `python -c '
...
 print(str(json.load(sys.stdin)["head"]["label"]).replace(":", " "))' </tmp/gh$$`
-rm /tmp/gh$$
 WHO=$1
 BRANCH=$2
+set -- `python -c '
+from __future__ import print_function
+import json, sys;
+print(str(json.load(sys.stdin)["head"]["repo"]["ssh_url"]))' </tmp/gh$$`
+REPO=$1
+rm /tmp/gh$$
```
mspncp: Why not make it a single python call which loads the json file once and prints all three arguments?

DDvO: I was thinking the same when I copied the Python code for simplicity.
I'm not a Python programmer so far, but meanwhile found how to do it nicely.

DDvO:
> While you're tweaking, would it be possible to have `$MERGE` default to "no"? The mere fact that this script defaults to squashing all commits into one has had me avoid it like the plague.

I was considering that as well but was reluctant because of the name of the tool.
Yet since there is strong consensus on changing the default I've done it now.

DDvO:
> A strong "yes" to defaulting `$MERGE` to no.

How about changing the name of the tool to `ghpush`?
This would reflect its semantics better, covering both the merge and rebase case.

review-tools/ghmerge
```diff
@@ -59,14 +59,12 @@ curl -s https://api.github.com/repos/openssl/$WHAT/pulls/$PRNUM >/tmp/gh$$
 set -- `python -c '
 from __future__ import print_function
 import json, sys;
-print(str(json.load(sys.stdin)["head"]["label"]).replace(":", " "))' </tmp/gh$$`
+input = json.load(sys.stdin)
+print(str(input["head"]["label"]).replace(":", " "),
+      str(input["head"]["repo"]["ssh_url"]))'        </tmp/gh3531`
```
mspncp: This is probably a leftover from testing? `/tmp/gh3531`

mspncp:
> > A strong "yes" to defaulting `$MERGE` to no.
> 
> How about changing the name of the tool to `ghpush`?
> This would reflect its semantics better, covering both the merge and rebase case.

I would leave the name as it is, because the main action is the merging (in the sense of integration) of the pull request into the master branch. That we integrate it linearly without creating an explicit merge commit is more a technical detail IMHO.

mspncp:
Essentially, it's a `git merge --ff-only`.

review-tools/README
```diff
@@ -115,6 +115,9 @@ else 'git merge --no-commit --squash'.
 It works on the current branch, which should be master or one of the stable
 releases. For the branch to be merged a suitable remote must be available.
 
+The tool accesses databases on api.github.com and www.openssl.org via HTTPS.
```
levitte: www.openssl.org is incorrect.

mspncp:
> > > A strong "yes" to defaulting `$MERGE` to no.
> > 
> > 
> > How about changing the name of the tool to `ghpush`?
> > This would reflect its semantics better, covering both the merge and rebase case.
> 
> I would leave the name as it is, because the main action is the merging (in the sense of integration) of the pull request into the master branch. That we integrate it linearly without creating an explicit merge commit is more a technical detail IMHO.

Looking at it once more, I think it's `$MERGE` which is misnamed, not `ghmerge`. I would suggest to name it `$SQUASH` and let it default to false.

review-tools/README
DDvO: So should I just remove "and www.openssl.org", 
or which other URLs are (potentially) contacted besides api.github.org?

review-tools/ghmerge
DDvO: oops, yes! Fixed.

DDvO:
> Looking at it once more, I think it's `$MERGE` which is misnamed, not `ghmerge`. I would suggest to name it `$SQUASH` and let it default to false.

I'm not so sure on this, since in both cases some (auto)squashing is done: 
`git merge --no-commit --squash $WORK`
vs.
`git rebase -i --autosquash`

review-tools/README
mspncp: Why not just write

    The tool accesses external databases via HTTPS (e.g., api.github.com).

since your main point is to inform the reader about the proxy settings?

DDvO:
Doing my most recent push using the new version of the tool,
I found it makes more sense to do the `git rebase -i --autosquash` before  `addrev` is called.
I've changed this and also added an initial `git pull` on master.

mspncp:
> > Looking at it once more, I think it's `$MERGE` which is misnamed, not `ghmerge`. I would suggest to name it `$SQUASH` and let it default to false.
> 
> I'm not so sure on this, since in both cases some (auto)squashing is done:
> `git merge --no-commit --squash $WORK`
> vs.
> `git rebase -i --autosquash`

It's a matter of how you view it: If the pull request has seven commits, each of which has several fixups, then the latter does a dumb squash-all-in-a-big-blob, while the former leaves the seven commits intact and only incorparates the fixups. But it's you to choose a name which suits you.

levitte:
> If the pull request has seven commits, each of which has several fixups, then the latter does a dumb squash-all-in-a-big-blob, while the former leaves the seven commits intact and only incorparates the fixups.

... but the other way around.  `--squash` is the "dumb squash-all-in-a-big-blob", while `--autosquash` "leaves the seven commits intact and only incorparates the fixups."

mspncp:
>... but the other way around.

Yeah, thanks for noticing. I mixed up the order.

DDvO:
> It's a matter of how you view it: If the pull request has seven commits, each of which has several fixups, then the former does a dumb squash-all-in-a-big-blob, while the latter leaves the seven commits intact and only incorporates the fixups.

[in this quote I've fixed the confusion mentioned above]

I know.

> But it's up to you to choose a name which suits you.

All right, so I've kept the variable name `MERGE`
but renamed the option '-nosquash' to the meanwhile more to-the-point '-autosquash'.

DDvO:
> Please reword the commit message of [e738545](https://github.com/openssl/tools/commit/e738545692ec2f960dbbfcaa09bb236eb9770138) to add an empty line between 'various tweaks' and the subitems

Done.

Unfortunately, such a change breaks the autosquash capability of the fixups for this commit because git references the base commit by its title (and not, e.g. by its hash-based ID).

In the future I'll make sure to put an empty line after the first line of multi-line commit messages.

mspncp:
>  git references the base commit by its title (and not, e.g. by its hash-based ID).

GIt does not reference the commit hash by design. Just imagine that git would do that. Then it would have to rewrite the commit messages of all fixups every time you rebase.
Fortunately, it's not such a big deal to reorder and edit the rebase-todo list manually, so an interactive rebase is still feasible :-)

review-tools/README
```diff
@@ -101,16 +105,37 @@ Checkout branch for query 145:
...
+
+--nomerge         Use 'git rebase -i --autosquash'.
+                  This is the default.
+--autosquash      An alias to '--nomerge'.
+--merge           Use 'git merge --no-commit --squash'.
+--squash          An alias to '--merge'.
+--nobuild         Do not use 'opensslbuild'.
+                  Else it is invoked with $CC defaulting to "ccache gcc".
+--tools           Select the 'tools' repository rather than 'openssl'.
+                  This implies '--merge' and '--nobuild'.
```
levitte: I would rather have `--autosquash` the default *everywhere*, at the very least for consistency.

paulidale: Agreed.  With almost everyone using fixup commits, it makes sense.

review-tools/README
DDvO: Done

review-tools/README
DDvO: Since `--autosquash` is, as written in the above addition to the README, an alias to `--nomerge` and `--nomerge` has recently become the default according to this PR, `--autosqash` is already the default.
Note also that for the alternative approach (triggered by `--merge` or  `--squash`) using `git merge --squash`, the `--autosquash` option is not supported by git.

DDvO:
I've just added  `--ff-only` to the `git merge` invocations to be on the safe side.

review-tools/ghmerge
```diff
@@ -113,12 +113,12 @@ git log $REL..
 git checkout $REL
 if [ "$MERGE" == "yes" ] ; then
     echo -n "Press return to merge --squash to $REL: "; read foo
-    git merge --no-commit --squash $WORK
+    git merge --ff-only --no-commit --squash $WORK
```
mspncp: 👍

review-tools/README
levitte: > `--autosqash` is already the default.

Then please explain to me the `MERGE=yes` in the lines quoted below.

``` shell
    --tools)
        WHAT=tools ; MERGE=yes ; BUILD=no ; shift
        ;;
    --web)
        WHAT=web ; MERGE=yes ; BUILD=no ; shift
        ;;
```

DDvO: Ah, now I understand what you meant. Right, whoever did those  `--tools` and `--web` options for some (strange) reason added `MERGE=yes` to them. 

I was already considering dropping `MERGE=yes` for them. Done now.

richsalz:
I suggest you avoid aliases, you don't need more than one way to do things.

t8m:
> I suggest you avoid aliases, you don't need more than one way to do things.

Yeah, especially because this is not a public-facing tool that we would need to keep backwards compatible.

DDvO:
> I suggest you avoid aliases, you don't need more than one way to do things.

I fully agree, but I'm surprised @richsalz that this suggestion comes from you.
I just went through the history of that tool and it turns out that essentially you introducted it,
including the aliases.

DDvO:
Anyway, which option names should survive?
`--merge` and `--nomerge` ?
`--squash`  and `--autosquash` ?
Or maybe `--merge` and `--autosquash` ?
Or anything new?

richsalz:
Yes, I wrote the tool.  (I wrote most of the things in this repo :).  I don't recall the aliases, but since you're doing a big cleanup, remove them.  Folks can always not install the current version.

DDvO:
No problem, but is there any preference on the names that should be kept?

Here's a further suggestion: `--merge` (implying squash) and `--rebase`  (implying autosquash).

t8m:
The relevant difference is only the --autosquash vs the forced --squash. I am actually not sure the --squash is really useful.

t8m:
It would be useful only if the PR's commit messages were really bogus and the committer would instead write a new message from scratch.
What might be useful is to have an option to skip the rebase -i --autosquash if the PR is clean and there are no fixup commits.

DDvO:
Thanks @t8m for sharing your thoughts.

So we could simply make autosquash the default,
`--squash` would become an alternative option, and all three other options can disappear.

DDvO:
Concerning the `rebase -i --autosquash`, in my view this does not hurt - the user will see the list of commits in his editor (which can be seen as opportunity for a sanity check) and if everything is nice and clean he can simply save&exit without changes.

mspncp:
> Thanks @t8m for sharing your thoughts.
> 
> So we could simply make autosquash the default,
> `--squash` would become an alternative option, and all three other options can disappear.

I agree with @t8m and @DDvO, but I would like to set the wording straight: The choice is not about the alternatives "autosquash" vs. "squash", it is about "interactive rebase" vs. "squash everything". 

The `--autosquash` option is just a comfort feature of the interactive rebase command. All git does is to reorder the items in the rebase-todo list and to replace some 'pick' actions by 'fixup' resp. 'squash' actions. So effectively it saves you some editing time, but the main point is that you are doing an interactive rebase.

And it makes a lot of sense that the interactive rebase is the default ghmerge action, whereas squashing requires an explicit `--squash` option.

DDvO:
Fine, so I've removed 3 merge/squash options and kept only non-interactive `--squash`
while interactive autosquash is the default.

DDvO:
I've meanwhile cleaned up up the (fixup) commits I've done so far.
All comments/requets have been addressed, also the formal change request by @mspncp.

DDvO:
I've done some further small improvements on `ghmerge` and its documentation.

I found that sometimes (after some more complex/messed commits that interfere with each other) the `--autosquash` option of interactive rebase re-arranges commits in a way that causes merge conflicts. Therefore I've fixed the cleanup behavior of `ghmerge` in this case and added the `--noautosquash` option that leaves out `--autosquash` on `git rebase -i`.

DDvO:
I've just added a further option to `ghmerge`:
using `--remote <arg>` one can select the git remote of the branch to pull from and push to.
This defaults to 'origin', the hitherto hard-wired value.

DDvO:
I've also slightly simplified the CLI of `addrev`:
Instead of `--prnum=NNN' one can now simply give `NNN`, e.g.:
```
addrev 1234 -2 steve @richsalz
```

mspncp:
> This defaults to 'origin', the hitherto hard-wired value.

Why not ask git instead of assuming 'origin'?

$ git remote -v | grep 'github.com/openssl/openssl'
github	https://github.com/openssl/openssl.git (fetch)
github	https://github.com/openssl/openssl.git (push)

mspncp:
Ups, wrong repo :-)

$ git remote -v | grep git.openssl.org
origin	git://git.openssl.org/openssl.git (fetch)
origin	openssl-git@git.openssl.org:openssl.git (push)

mspncp:
A classical use-case for good'ol awk:

$ git remote -v | awk '/git.openssl.org.*(push)/{ print $1; }'
origin

DDvO:
> $ git remote -v | awk '/git.openssl.org.*(push)/{ print $1; }'
> origin

@mspncp,  good thought and thanks for the code snippet. 
I've just extended it by ` | head -n 1` for the case that more than just one such remote is listed.
I've updated also the README in this PR accordingly. 
Again all your change requests are fulfilled.

DDvO:
Trying the `--squash` option of the latest `ghmerge` on #11134 yesterday revealed that its error handling still deserved some improvements, which in turn motivated a further tweak also on `addrev`. These are done in the two most recent commits.

DDvO:
Any further comments or requests for changes?
Else would be nice if this can approved such that it becomes directly usable.

DDvO:
Thanks @t8m and @paulidale for your swift approvals.

Can an OMC member please now do the merge?
You may already take advantage of the new `ghmerge` to merge itself :wink:, e.g., like this:
```
ghmerge --tools 59 @t8m paulidale
```

DDvO:
Thanks @t8m and @paulidale for your swift approvals.

Can an OMC member please now do the merge?
You may already take advantage of the new `ghmerge` to merge itself :wink:, e.g., like this:
```
ghmerge --tools 59 @t8m paulidale
```

paulidale:
I've merged it but I don't think I used the new ghmerge.  I would have used the new versions of things it calls.

t8m:
> I've merged it but I don't think I used the new ghmerge. I would have used the new versions of things it calls.

Well now after this merge the ghmerge is becoming really useful tool.

DDvO:
> Well now after this merge the ghmerge is becoming really useful tool.

Very pleased to hear!

PR #60, opened 2020-02-12
Add the ability to cope with a raw token for authentication (if our

### LINKED ISSUE #53: Notifications on "urgent" labels
Addition of labels doesn't result in a notification being sent out. It would be good if we sent out a notification when an urgent label has been applied to a PR

iamamoose:
file contains just a hash assume it's a token)

Add the ability to comment on urgent PR
Fixes #53

Don't move things to 'ready to merge' until CI has passed


levitte:
Same comment here as in #62

t8m:
In different repo now.

PR #62, opened 2020-03-09
More Notifications

iamamoose:
Add commenting to issues past 24 hours but that we are not automatically changing the label for.  Make sure we don't keep doing this every time we run by looking for comments made by our openssl-machine account


levitte:
Er, was this ever merged, @iamamoose?  This has moved to https://github.com/openssl/omc-tools.  I'd say that you can merge directly to that repo, or recreate this PR there.  Your choice.

levitte:
There's a conflict, so please rebase and get it re-approved so we can actually merge it!

(we're being quite sloppy with the tools repo)

t8m:
In different repo now.

PR #65, opened 2020-04-21
Add a note about closing the github advisory

mattcaswell:
If we've used a private github fork then we should close the advisory
and fork after the release is done.


mattcaswell:
Pushed.

PR #67, opened 2020-05-25
Improve ghmerge, gitaddrev, and opensslbuild

DDvO:
- Add check of `REMOTE` and add `--verbose` option to `ghmerge`
- Improve author matching for skipping author in list of reviewers in `gitaddrev`

### DIFF
--- a/review-tools/ghmerge
+++ b/review-tools/ghmerge
@@ -2,10 +2,8 @@
 
 set -o errexit
 
-VERBOSE=""
 WHAT=openssl
 BUILD=yes
-TRIVIAL=""
 INTERACTIVE=yes
 AUTOSQUASH="--autosquash"
 [ -z ${CC+x} ] && CC="ccache gcc" # the default otherwise is "ccache clang-3.6"
@@ -20,20 +18,15 @@ if [ ! -d .git ] ; then
     exit 1
 fi
 
+ADDREVOPTS=""
 # Parse JCL.
 while true ; do
     case "$1" in
-    --verbose)
-        VERBOSE="--verbose" ; shift
-        ;;
     --tools)
         WHAT=tools ; BUILD=no ; shift
         ;;
     --web)
         WHAT=web ; BUILD=no ; shift
-        ;;
-    --trivial)
-        TRIVIAL="--trivial" ; shift
         ;;
     --noautosquash)
         AUTOSQUASH="" ; shift
@@ -55,9 +48,10 @@ while true ; do
         shift
         break
         ;;
-    -*)
-        echo "$0: Unknown flag $1"
-        exit 1
+    -*) # e.g., --verbose, --trivial, --myemail=...
+        ADDREVOPTS="$ADDREVOPTS $1"
+        shift
+        break
         ;;
     *)
         break
@@ -66,7 +60,7 @@ while true ; do
 done
 
 if [ $# -lt 2 ] ; then
-    echo "Usage: $0 [flags] prnum reviewer..."
+    echo "Usage: $0 [flags, including addrev flags] prnum reviewer..."
     exit 1
 fi
 PRNUM=$1 ; shift
@@ -122,7 +116,7 @@ git diff $REL
 if [ "$INTERACTIVE" == "yes" ] ; then
     echo -n "Press Enter to interactively rebase $AUTOSQUASH $BRANCH on $REL: "; read foo
     git rebase -i $AUTOSQUASH $REL || (git rebase --abort; exit 1)
-    addrev $VERBOSE $TRIVIAL --prnum=$PRNUM $TEAM ${REL}..
+    addrev $ADDREVOPTS --prnum=$PRNUM $TEAM ${REL}..
 fi
 
 echo Log since $REL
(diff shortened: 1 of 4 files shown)


DDvO:
Meanwhile I've added two commits containing 
a couple of minor further improvements to `ghmerge` and `opensslbuild`

DDvO:
> Still LGTM.

> I think another OMC reviewer is required.

Thanks @paulidale for reviewing those recent minor additions and re-confirming.

Since the 2nd approval has still been missing anyway, 
I've just added two more minor improvements to this PR.

DDvO:
Ping for 2nd OTC approval

review-tools/ghmerge
```diff
@@ -24,14 +28,11 @@ while true ; do
...
         ;;
-    --trivial)
-        TRIVIAL="--trivial" ; shift
-        ;;
     --noautosquash)
         AUTOSQUASH="" ; shift
         ;;
     --squash)
-        INTERACTIVE=no ; shift
+        INTERACTIVEno ; shift
```
t8m, line 35: a typo

DDvO: Oops, thanks, fixed.

DDvO:
Thanks @paulidale and @t8m for having a (repeated) look at this and approving it!

DDvO:
Since unfortunately normal committers like me cannot push to the tools repository (which is something that does not really make sense IMHO),
I ask any OTC member to do the merge.

t8m:
oops, I've misclicked

t8m:
So apparently OTC members cannot push there either. We need an OMC member to do that. 
@openssl/omc please can you merge?

The merge I've inadvertently done on github will be overridden once the proper merge on the base repository is done.

DDvO:
> So apparently OTC members cannot push there either. We need an OMC member to do that.

This reminds me of the famous words from George Orwell's novella "Animal Farm":

> All animals are equal, but some animals are more equal than others.

mattcaswell:
> So apparently OTC members cannot push there either. We need an OMC member to do that.
@openssl/omc please can you merge?

OTC members should be able to do this...not sure why they can't.

DDvO:
> OTC members should be able to do this...not sure why they can't.

Maybe because some OTC members, quoting again Orwell, are more equal than others :wink:

mattcaswell:
I think @levitte may have fixed this...@t8m can you try again?

DDvO:
BTW, what is the rationale why no (normal) committer is allowed to push to the tools repo?

mattcaswell:
> BTW, what is the rationale why no (normal) committer is allowed to push to the tools repo?

Originally this was an OMC only repo. We recently reviewed that since most of the contents are inherently technical in nature - so we made it OTC owned instead. I'm not sure why we didn't just make the review rules the same as for the main source repo. Possibly because we just didn't consider it. The OTC could still choose to make that change.

DDvO:
Thanks @mattcaswell for the explanation!

t8m:
> I think @levitte may have fixed this...@t8m can you try again?

Yeah, it worked now. So the proper commits from the base repo are now in there.

DDvO:
Thanks @levitte and @t8m

review-tools/opensslbuild
```diff
@@ -46,11 +46,11 @@ grep -q doc-nits Makefile && make doc-nits
...
-echo Making
+echo Making build
 make -s -j20
 make -s -j20 build_tests >/dev/null
 s=$?
 test $execonly = yes && exit $s
 
 echo Making test
-make test
+make -j4 test
```
romen, line 56: @DDvO why not `-j20` like the build targets above?

In the end the tests are not really going to run in parallel (at least in `master`)

~~~Makefile
# Excerpts from the generated Makefile on `master`
# [...]

# Kept around for backward compatibility
build_apps build_tests: build_programs

# [...]

test: tests
tests: build_generated
        $(MAKE) depend && $(MAKE) _tests
_tests: build_programs_nodep build_modules_nodep link-utils
        @ :
        ( SRCTOP=$(SRCDIR) \
          BLDTOP=$(BLDDIR) \
          PERL="$(PERL)" \
          EXE_EXT= \
          $(PERL) $(SRCDIR)/test/run_tests.pl $(TESTS) )
        @ :

list-tests:
        @ :
        @SRCTOP="$(SRCDIR)" \
         $(PERL) $(SRCDIR)/test/run_tests.pl list
        @ :

~~~

As far as I can tell the `_tests` target spawns a single `perl` process, so giving `make` more than one job runner is not really going to run the tests in parallel (even then `make tests` runs a separate `make _tests` process, to which `-j4` is not passed anyway): the only thing running in in parallel is the build of the `build_generated` dependency (and not the {`build_programs_nodep`, `build_modules_nodep`, `link-utils`} dependencies of `_tests`, as `_tests` is run in a single job make instance).


---

Side note: I once looked into trying to give `run_tests.pl` the capability of running the tests in parallel, but my Perl-fu is too weak!

Do you have plans to look into that?

DDvO: > @DDvO why not `-j20` like the build targets above?

Well, simply because I typically use `make` with `-j4` and I suspect that a larger number would not help for most CPUs.

> In the end the tests are not really going to run in parallel (at least in `master`)

Yes, but test prerequisites could benefit from potential parallel execution.
Though this likely only holds for any build steps, but in this case they will have already been done by the preceding `make` invocations.

> As far as I can tell the `_tests` target spawns a single `perl` process, so giving `make` more than one job runner is not really going to run the tests in parallel. [...]

Unfortunately, yes.

> Side note: I once looked into trying to give `run_tests.pl` the capability of running the tests in parallel, but my Perl-fu is too weak!
> 
> Do you have plans to look into that?

I don't have experience how to do that either and do not have plans for doing that.

Of course, would be really nice if tests were considerably speeded up by (somewhat) parallel execution.
I suggest opening an issue requesting that feature.

DDvO: Thinking a little more about this, there is actually no need for Perl magic for achieving this!
We can simply take advantage of the `make` capabilities for running tasks in parallel,
by partitioning in the `Makefile` the list of tests to run into chunks of about equal run time.

romen: @DDvO less magic than I expected seemed to be required: https://github.com/openssl/openssl/pull/12326

PR #68, opened 2020-06-09
rand-drbg: Add conversion script to create data for DRBG testing.

paulidale:
NIST's DRBG test data set isn't ideal for evp_test to process.  This script massages the data into a format which is more suitable.

### DIFF
--- a/nist-conversion/README.md
+++ b/nist-conversion/README.md
@@ -1,0 +1,6 @@
+This directory contains tools that are used to convert between NIST supplied
+test data sets and internal formats.
+
+The `convert_nist_drbg_test_data.lua` script converts the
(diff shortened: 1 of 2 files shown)


paulidale:
OMC review required (I think).

richsalz:
FWIW, I don't think we need to include the ZIP file (and since it's so big, shouldn't).
Ideally, the conversion script is rewritten in perl.

paulidale:
I don't mind dropping the zip file.  I included it because it represents the data at the point in time they were snapshot.

paulidale:
Ping @openssl/omc

paulidale:
Ping @openssl/omc

mattcaswell:
> OMC review required (I think).

Actually OTC approval is needed for the tools repo.

paulidale:
Ping @openssl/otc

paulidale:
Merged, thanks.

PR #71, opened 2020-07-02
[run-checker]: default to 4 parallel test jobs

romen:
<https://github.com/openssl/openssl/pull/12326> added support for
running the tests in parallel. The default is still to run them
sequentially.

This alters run-checker so that it defaults to 4 test jobs in parallel,
which seems a safe assumption given we are already using 4 building
jobs.


levitte:
Wow, this had been sitting a while!

Merged

0cce4862d4ea91d81b6ebcfdc3106d1331f2f291 [run-checker]: default to 4 parallel test jobs

PR #72, opened 2020-10-08
Many small improvements of ghmerge and addrev

DDvO:
addrev: 
* Improve flexibility providing reviewer names vs. commit refs (filter args) 
* Add warning on overriding filter args

ghmerge:
* Improve flexibility providing reviewer names vs. commit refs (gitaddrev filter args)
* Add `--cherry-pick` option (as alternative to rebase), useful for back-porting
* Add `--ref` option to checkout given reference at first
* Unify behavior on error in CLI arguments and improve usage output
* Make sure that on build error all error output is shown and then the tool exits
* Improve diagnostics on cleanup, distinguishing failure and non-error exit
* Various small improvements on user interaction, extending messages etc.
* Various minor code cleanups

### DIFF
--- a/review-tools/addrev
+++ b/review-tools/addrev
@@ -4,7 +4,7 @@ use strict;
 use warnings;
 
 my $args = "";
-my $filterargs = "HEAD^..";
+my $filterargs = "";
 my $list_reviewers = 0;
 my $help = 0;
 my $haveprnum = 0;
@@ -18,7 +18,8 @@ foreach (@ARGV) {
         $args .= "--reviewer=$_ ";
     } elsif (/^[-\w]+$/) {
         if (/^[0-9a-f]{7,}+/) {
-            $filterargs = "$_";
+            print "Warning: overriding previous filter args $filterargs\n" if $filterargs ne "";
+            $filterargs = $_;
         } else {
             $args .= "--reviewer=$_ ";
         }
@@ -46,6 +47,7 @@ foreach (@ARGV) {
     } elsif (/^--commit=(.+)$/) {
         $args .= "--commit=$1 ";
     } elsif (/^-(\d+)$/) {
+        print "Warning: overriding previous filter args $filterargs\n" if $filterargs ne "";
         $filterargs = "HEAD~$1..";
     } elsif (/^--list$/) {
 	$list_reviewers = 1;
@@ -54,9 +56,11 @@ foreach (@ARGV) {
 	$help = 1;
 	last;
     } else {
+        print "Warning: overriding previous filter args $filterargs\n" if $filterargs ne "";
         $filterargs = $_;
     }
 }
+$filterargs = "HEAD^.." if $filterargs eq "";
 
 if ($list_reviewers) {
     system("gitaddrev --list");
--- a/review-tools/addrev
+++ b/review-tools/addrev
@@ -14,8 +14,14 @@ my $useself = 1;
 my $my_email;
 
 foreach (@ARGV) {
-    if (/^[a-z]+$/ || /^\@\w(?:[-\w]*\w)?$/) {
+    if (/^\@.+$/) {
         $args .= "--reviewer=$_ ";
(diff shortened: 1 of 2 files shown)


DDvO:
As a Committer I cannot set any flag such as `otc review pending` for PRs like this in the `tools` repo.

t8m (approved):
LGTM

A second review by someone who uses ghmerge/addrev would be nice.

DDvO:
> LGTM
> 
> A second review by someone who also uses ghmerge/addrev would be nice.

Thanks a lot @t8m for picking up this PR, reviewing and approving it!

Meanwhile I had done a tweak to one of my commits, removing again the need to press enter just before `addrev` is called:
```
-    echo -n "Press Enter to call addrev $ADDREVOPTS --prnum=$PRNUM $TEAM ${REL}.."; read foo
+    echo "Calling addrev $ADDREVOPTS --prnum=$PRNUM $TEAM ${REL}.."
```
I've just pushed this fixup commit and ask you to reconfirm the approval.

t8m:
Still approved.

paulidale (commented):
Is it time to document how these scripts work?
Or at least some sample usages?

DDvO:
> Is it time to document how these scripts work?
> Or at least some sample usages?

Good point. This should have been done right from the beginning, as done for `addrev`.
I've just added an explicit `--help` option and significantly extended the usage hints (now describing options and giving examples).

paulidale (commented):
The help option is good.  Thanks.

DDvO:
> The help option is good. Thanks.

Pleased to hear. 
Can some OTC member please take over and merge this?
Even with approvals, as a normal Committer I won't be able to.

paulidale (approved):

paulidale:
Merged.

DDvO:
Thanks @paulidale and @t8m!

mattcaswell:
Something in this PR seems to have broken addrev's ability to handle multiple commits. In the case of PR 13368 I have 17 commits I need to add a header to:

````
$ addrev -17 --prnum=13368 levitte
Rewrite d3d2c0dc68e6eebbfe7b1fc2ea653225a9f37a94 (1/1) (0 seconds passed, remaining 0 predicted)    Unknown reviewers: -17
msg filter failed: gitaddrev --reviewer=-17 --prnum=13368 --reviewer=levitte --myemail=matt@openssl.org
 
addrev failed
Died at /home/matt/dev/tools/review-tools/addrev line 86.
````

**Comment by t8m** [comment]
#76 should help

**Comment by DDvO** [comment]
> `addrev -17 --prnum=13368 levitte`

@mattcaswell, I wonder why you use an explicit number of commits.

I thinkg typically it should suffice to give the commit range `master..` (so no need to manually count commits)
or use any other branch ref (`ghmerge` uses `master` by default):

`addrev --prnum=13368 levitte master..`

which BTW is more conveniently achieved by

`ghmerge --nobuild 13368 levitte`

**Comment by levitte** [comment]
> @mattcaswell, I wonder why you use an explicit number of commits.

It's his work flow, and it was a supported way.  Either way, `-17` is meaningless for revisions per se.

In another iteration, it's possible that we should make better use of `git rev-parse`...  I've done some experiments which were inconclusive, so I wouldn't recommend jumping at the opportunity too fast.

**Comment by DDvO** [comment]
> > @mattcaswell, I wonder why you use an explicit number of commits.
> 
> It's his work flow, and it was a supported way. 

I did not want to question that this should be supported,
but just wanted to hint for convenience that it's likely not needed this way.


> In another iteration, it's possible that we should make better use of `git rev-parse`... I've done some experiments which were inconclusive, so I wouldn't recommend jumping at the opportunity too fast.

When using `ghmerge`, the default ref (usually, `master`) is determined like this:
```
REF=`git rev-parse --abbrev-ref HEAD`
```

DDvO:
... and then the default commit range will be `master..` (or `REF..` with whatever other REF chosen explicitly or implicitly).

PR #76, opened 2020-11-18
Fix regression from latest addrev change

t8m:


t8m:
@mattcaswell this should fix the regression you see.

mattcaswell (approved):

t8m:
Merged

review-tools/addrev
```diff
@@ -16,7 +16,7 @@ my $my_email;
 foreach (@ARGV) {
     if (/^\@.+$/) {
         $args .= "--reviewer=$_ ";
-    } elsif (/^[-\w]+$/) {
+    } elsif (/^\w[-\w]+$/) {
```
levitte, line 19: Side note: one character reviewer names are no longer allowed :wink:

DDvO:
Thanks @t8m for quickly providing a fix.

review-tools/addrev
DDvO: > Side note: one character reviewer names are no longer allowed :wink:

o :wink:

So it would have been better to use `/^\w[-\w]*$/`

DDvO: I'm going to fix this very minor issue if /when doing any further improvements of `addrev`.

PR #79, opened 2020-12-02
Improve ghmerge ref handling and documentation

DDvO:
* `ghmerge`: Make pulling the latest REMOTE/REF work als for non-default REF
    
    This avoids potential git errors such as:
    
    You asked to pull from the remote 'upstream', but did not specify
    a branch. Because this is not the default configured remote
    for your current branch, you must specify a branch on the command line.

* `ghmerge`: improve doc of `--remote` and `--ref` options

* `addrev`: re-enable use of singe-character review names

### DIFF
--- a/review-tools/addrev
+++ b/review-tools/addrev
@@ -16,7 +16,7 @@ my $my_email;
 foreach (@ARGV) {
     if (/^(--prnum=)?(\d{1,6}+)$/) {
         $args .= "--prnum=$2 ";
-	$haveprnum = 1;
+        $haveprnum = 1;
     } elsif (/^\@.+$/) {
         $args .= "--reviewer=$_ ";
     } elsif (/^\w[-\w]*$/) {
@@ -43,18 +43,18 @@ foreach (@ARGV) {
     } elsif (/^--myemail=(.+)$/) {
         $my_email = $1;
     } elsif (/^--nopr$/) {
-	$haveprnum = 1;
+        $haveprnum = 1;
     } elsif (/^--commit=(.+)$/) {
         $args .= "--commit=$1 ";
     } elsif (/^-(\d+)$/) {
         print "Warning: overriding previous filter args $filterargs\n" if $filterargs ne "";
         $filterargs = "HEAD~$1..";
     } elsif (/^--list$/) {
-	$list_reviewers = 1;
-	last;
+        $list_reviewers = 1;
+        last;
     } elsif (/^--help$/ || /^-h$/) {
-	$help = 1;
-	last;
+        $help = 1;
+        last;
     } else {
         print "Warning: overriding previous filter args $filterargs\n" if $filterargs ne "";
         $filterargs = $_;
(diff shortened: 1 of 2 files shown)


t8m (approved):
LGTM

DDvO:
> LGTM

Thanks @t8m for having a look and approving.

I just found that there was also an issue with using `addrev` with giving PR numbers without the (optional) `--prnum=` prefix.
So I've added a commit fixing this:
* `addrev`: Fix regression on parsing bare prnum; correct help output

review-tools/addrev
```diff
@@ -14,7 +14,10 @@ my $useself = 1;
 my $my_email;
 
 foreach (@ARGV) {
-    if (/^\@.+$/) {
+    if (/^(--prnum=)?(\d{1,6}+)$/) {
+        $args .= "--prnum=$2 ";
+	$haveprnum = 1;
```
t8m, line 19: indent by 2 more spaces?

review-tools/addrev
```diff
@@ -14,7 +14,10 @@ my $useself = 1;
 my $my_email;
 
 foreach (@ARGV) {
-    if (/^\@.+$/) {
+    if (/^(--prnum=)?(\d{1,6}+)$/) {
```
t8m, line 17: I understand that you want to support all numeric 7 character git commit id. Hopefully the PR numbers won't be in millions range too soon. :grin:

t8m (approved):

review-tools/addrev
DDvO: This was due to a tab char. 
It turns out there were many of them in that file, which I've just eliminated.

review-tools/addrev
DDvO: Yes. Recently the PR numbers became 5-digit, and it will certainly take a while until 6 digits are needed
By that time likely also more than 7 hex digits will be needed to identify commits,
so then our (grand)children can simultaneously increment both limits by one :sweat_smile:

t8m (approved):

DDvO:
I think a 2nd approval is needed here, and I cannot set any labels to indicate that...

t8m:
I don't know if it is needed. Do there have to be two OTC member approvals on tools? I did not think so.

paulidale (approved):

newblaque (approved):

DDvO:
Can an OTC member please merge this?

BTW, would be nice if committers got privilege for that, too.

t8m:
Merged now.

t8m:
Aargh. I've used --tools with ghmerge, but apparently it did not pass the option to addrev so the references in the commits are wrong. Unfortunately it is already pushed.

DDvO:
Good spotting.
I just found this was not the first and only glitch like that:
```
git log|grep github.com/openssl/openssl|wc
```
on the tools master gives 50 hits since 2017.

Yet another item to improve on `ghmerge` :)
I never came across this because i cannot merge to the tools repo.
Since I cannot really test it, I'm leaving this bug for someone else to fix.

t8m:
#80

PR #85, opened 2021-05-01
run-checker: add more builds with FIPS enabled and other disabled.

paulidale:
Not all of these work yet, this won't be merged until they do.

### DIFF
--- a/run-checker/run-checker.sh
+++ b/run-checker/run-checker.sh
@@ -46,7 +46,13 @@ no-tls1_2-method no-dtls1-method no-dtls1_2-method no-siphash no-tls1_3 no-sm2
 no-sm3 no-sm4 enable-trace no-legacy no-cached-fetch no-autoload-config
 'no-buildtest-c++' no-bulk no-cmp no-ktls no-module no-padlockeng
 no-pinshared no-secure-memory no-siv no-uplink enable-acvp-tests enable-fips
-'enable-fips no-fips-securitychecks'
+'enable-fips no-fips-securitychecks' 'enable-fips enable-acvp-tests'
+'enable-fips no-cmac' 'enable-fips no-cmac' 'enable-fips no-des'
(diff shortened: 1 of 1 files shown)


paulidale:
The FIPS no-des fix is in #15105, the rest seem good locally.

mattcaswell (approved):

paulidale:
Added a fixup to remove the duplicate entry.  Reapproval required.

run-checker/run-checker.sh
```diff
@@ -46,7 +46,13 @@ no-tls1_2-method no-dtls1-method no-dtls1_2-method no-siphash no-tls1_3 no-sm2
 no-sm3 no-sm4 enable-trace no-legacy no-cached-fetch no-autoload-config
 'no-buildtest-c++' no-bulk no-cmp no-ktls no-module no-padlockeng
 no-pinshared no-secure-memory no-siv no-uplink enable-acvp-tests enable-fips
-'enable-fips no-fips-securitychecks'
+'enable-fips no-fips-securitychecks' 'enable-fips enable-acvp-tests'
+'enable-fips no-cmac' 'enable-fips no-cmac' 'enable-fips no-des'
```
paulidale, line 50: On second thoughts are the FIPS + disabled algorithms actually meaningful?   It seems unlikely that the FIPS provider could be built in an approved manner if they are excluded.

t8m: Well, somebody could do a private validation with FIPS and disabled algorithms. But I am not sure we need to worry about such in runchecker builds.

paulidale: I've pulled the FIPS + disabled algorithms options.

t8m (approved):

paulidale:
Merged to master.  Let the fun begin.
Thanks for all of the helpful comments.

PR #87, opened 2021-05-12
run-checker: reduce the number of builds

paulidale:
With the addition of most run-checker jobs to GitHub Actions, there is no need to continue running these jobs ourselves.

This won't be merged until after openssl/openssl#15129 is.


mattcaswell (approved):

paulidale:
Merged, thanks for the review.

PR #89, opened 2021-05-21
Send release emails from the owner of the signing key

mattcaswell:
Announcement emails should be sent from the email account of the owner of
the signing key, otherwise some email clients will fail to verify the key
correctly.

A longer term solution will be to have a separate release signing key.


paulidale (approved):
LGTM

levitte (approved):

levitte:
We do have an OpenSSL team key...

mattcaswell:
Pushed.

PR #91, opened 2021-08-27
run-checker: remove the fuzzing builds

paulidale:
Not to be merged until after openssl/openssl#16438 is merged and working.

### DIFF
--- a/run-checker/run-checker.sh
+++ b/run-checker/run-checker.sh
@@ -21,9 +21,7 @@
 # hook-takedown - called at the very end
 
 here=$(cd $(dirname $0); pwd)
-opts=( ''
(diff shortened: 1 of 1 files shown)


paulidale:
Ping for review openssl/openssl#16438 merges soon...

levitte (approved):

levitte:
This means that we can drop the run-checker VM as well, right?

paulidale:
Yes, but separate PR

t8m (approved):

t8m:
Would it actually make sense to remove the run-checker directory completely?

paulidale:
Yes we should remove it, in a different PR.

t8m:
Merged. Thank you.

PR #94, opened 2021-10-07
ghmerge: Fix branch handling

DDvO:
* Correct saving and restoring original branch state
* Properly catch the error that `copy-of-...` already exists
* Improve robustness and clarity regarding which ref branch is used
* Avoid checking out ref branch if not needed

PR #95, opened 2021-10-23
Improve `pick-to-branch`

DDvO:
This has been motivated by https://github.com/openssl/tools/pull/94#issuecomment-939163365.
I found `pick-to-branch` rather cumbersome to use (as opposed to `ghmerge`),
but anyway made a quick pass over it, doing the following improvements:

* Improve error handling, as suggested in https://github.com/openssl/tools/pull/94#issuecomment-939163365
* Preserve current branch and its state if it is not the target
* Fix the case that commit id is derived from HEAD of master
* pick-to-branch: Improve user guidance

### DIFF
--- a/review-tools/pick-to-branch
+++ b/review-tools/pick-to-branch
@@ -16,7 +16,7 @@ case $# in
     b=$2
     ;;
 1)
-    id=`git branch -v | awk '$2=="master" { print $3; }'`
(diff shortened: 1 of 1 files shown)


DDvO:
Also this can be merged.

t8m (approved):

t8m:
Merged to master. Thank you

PR #96, opened 2021-11-05
pick-to-branch: fix inferring the id from master branch

t8m:
The original way did not account for master being the currently active branch.

### DIFF
--- a/review-tools/pick-to-branch
+++ b/review-tools/pick-to-branch
@@ -6,7 +6,7 @@ function usage {
     If this is not the current branch, the current branch and its state are preserved.
 
     The commit can be given in the form of a branch name.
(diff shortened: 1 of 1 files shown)


paulidale (approved):

DDvO (approved):
I do not understand what was wrong with the original way, 
but the new one works at least equally well and is more direct/efficient.
I also like the slightly further improved diagnostics.

levitte:
> I do not understand what was wrong with the original way,

When master is the current branch, the `git branch -v` output is this:

```
* master                                                   884400d789 ...
```

In other words, the `awk` scriplet tries to match `*` with `master` in that case.

Using `git show` instead is much more script friendly.

t8m:

PR #97, opened 2021-11-08
ghmerge: Fix behavior on failed cherry-pick

DDvO:
In case the cherry-pick failed, clean up the state
(as meanwhile done also by `pick-to-branch`).

### DIFF
--- a/review-tools/ghmerge
+++ b/review-tools/ghmerge
@@ -11,7 +11,8 @@ Option style arguments:
 --tools             Merge a tools PR (rather than openssl PR)
 --web               Merge a web PR (rather than openssl PR)
 --remote <remote>   Repo to merge with (rather than git.openssl.org), usually 'upstream'
---ref <branch>      Branch to merge with (rather than current branch), usually 'master'
+--target <branch>   Merge target (rather than current branch), usually 'master'
+--ref <branch>      A synonym for --target
 --cherry-pick [<n>] Cherry-pick the last n (1 <= n <= 99, default: 1) commits, rather than rebasing
 --squash            Squash new commits non-interactively (allows editing msg)
 --noautosquash      Do not automatically squash fixups in interactive rebase
@@ -22,8 +23,8 @@ Examples:
   ghmerge 12345 mattcaswell
   ghmerge 12345 paulidale t8m --nobuild --myemail=dev@ddvo.net
   ghmerge edd05b7^^^^..19692bb2c32 --squash -- 12345 levitte
-  ghmerge 12345 slontis --ref openssl-3.0
-  ghmerge --nobuild --ref openssl-3.0 --cherry-pick 3 16051 paulidale"
+  ghmerge 12345 slontis --target openssl-3.0
+  ghmerge --nobuild --target openssl-3.0 --cherry-pick 3 16051 paulidale"
     exit 9
 }
 
@@ -34,7 +35,7 @@ PICK=no
 INTERACTIVE=yes
 AUTOSQUASH="--autosquash"
 REMOTE=""
-REF=""
+TARGET=""
 BUILD=yes
 [ -z ${CC+x} ] && CC="ccache gcc" # opensslbuild will otherwise use "ccache clang-3.6"
 
@@ -82,12 +83,12 @@ while [ $# -ne 0 ]; do
         fi
         shift; REMOTE=$1; shift
         ;;
-    --ref)
+    --target|--ref)
         if [ $# -lt 2 ] ; then
             echo "Missing argument of '$1'"
             usage_exit
         fi
-        shift; REF=$1; shift
+        shift; TARGET=$1; shift
         ;;
     --)
         if [ $# -lt 3 ] ; then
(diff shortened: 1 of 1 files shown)


t8m (approved):

paulidale (approved):

DDvO:
I meanwhile found that the solution was not sufficient (namely, for situations where the user aborts cherry-picking etc. using Ctrl-C), so I extended it.

DDvO:
On this occasion extended `ghmerge --cherry-pick` to picking potentially more than one commit.
With this generalization, one can nicely use `ghmerge` also for backporting PRs with more than one commit.

For instance, after
```
ghmerge --nobuild 16051 paulidale
```
I just used successfully
```
ghmerge --nobuild --ref openssl-3.0 --cherry-pick 3 16051 paulidale
```

paulidale (approved):

review-tools/ghmerge
```diff
@@ -12,7 +12,7 @@ Option style arguments:
 --web               Merge a web PR (rather than openssl PR)
 --remote <remote>   Repo to merge with (rather than git.openssl.org), usually 'upstream'
 --ref <branch>      Branch to merge with (rather than current branch), usually 'master'
---cherry-pick       Use cherry-pick (rather than pull --rebase)
+--cherry-pick <n>   Cherry-pick the last n commits (rather than rebasing)
```
t8m, line 15: If anyone was using --cherry-pick without the <n> parameter they will have to start putting it in there. I do not know if we care about that. @mattcaswell @paulidale @levitte

DDvO: I could also make `n=1` the default if `--cherry-pick` is not followed by number (i.e., sequence of decimal digits).

t8m: > I could also make `n=1` the default if `--cherry-pick` is not followed by number (i.e., sequence of decimal digits).

Good idea.

t8m: Perhaps even if the number after it is larger than 20 or something

t8m: `ghmerge --nobuild --ref openssl-3.0 --cherry-pick 16051 paulidale` - otherwise this won't work.

DDvO: Yeah, even better to limit `n` in order to prevent clashes with any given commit ID.
Done so, allowing (to be on the safe side) `1 <= n <= 99`.

review-tools/ghmerge
```diff
@@ -60,7 +60,11 @@ while [ $# -ne 0 ]; do
         WHAT=web ; BUILD=no ; shift
         ;;
     --cherry-pick)
-        shift; PICK=$1; shift
+        shift;
+        PICK=1;
+        if [ $1 != "" ] && [ $1 -ge 1 ] && [ $1 -le 99 ]; then
```
t8m, line 65: This will probably emit errors if the argument after --cherry-pick is a non-numeric

t8m: You can add 2>/dev/null to the test statements  with -ge and -le to eliminate this.

t8m: Actually it would be sufficient to add it to the -ge test call as the other won't be run.

DDvO: Right; fixed.

review-tools/ghmerge
```diff
@@ -59,7 +60,11 @@ while [ $# -ne 0 ]; do
         WHAT=web ; BUILD=no ; shift
         ;;
     --cherry-pick)
-        PICK=yes ; shift
+        shift;
+        PICK=1;
+        if [ "$1" != "" ] && [ $1 -ge 1 ] && [ $1 -le 99 ]; then
+            PICK=$1 ; shift
+        fi
```
t8m, line 67: you need to do:
`if [ "$1" != "" ] && [ $1 -ge 1 ] 2>/dev/null && [ $1 -le 99 ]; then`

DDvO: Ah, I had not seen your hint regarding `2>/dev/null`, done now.

t8m (approved):

paulidale (approved):

t8m (approved):

DDvO:
Thanks @t8m for re-approving.
Can this already be merged as well?

t8m:
Merged. Thank you.

DDvO:
After these further improvements of `ghmerge` and 
the generalization of its `--cherry-pick` capabilities,
I actually see little remaining use of `pick-to-branch`.

t8m:
I disagree. The pick-to-branch seems still very useful to me. I do not see the --cherry-pick option to be a real replacement as that still does addrev and stuff, which is useless if I am just picking commits that were recently merged to master. Please see my PR on pick-to-branch to allow multiple commits to be cherry-picked.

DDvO:
> I disagree. The pick-to-branch seems still very useful to me. I do not see the --cherry-pick option to be a real replacement as that still does addrev and stuff, which is useless if I am just picking commits that were recently merged to master.

Right, the `addrev` stuff does not need to be re-done when backporting a PR.
Still I find `ghmerge` more convenient to use in this case.
BTW, I defined shell aliases like this:
```
alias ghm='ghmerge --nobuild --target master'
alias ghm3='ghmerge --nobuild --target openssl-3.0 --cherry-pick'
alias ghm1='ghmerge --nobuild --target OpenSSL_1_1_1-stable --cherry-pick'
```
such that, e.g., after a
```
ghm 17056 levitte t8m paulidale
```
I can simply recall the last shell command and adapt it like this:
```
ghm3 2 17056 levitte t8m paulidale
```
and am done :-)

> Please see my PR on pick-to-branch to allow multiple commits to be cherry-picked.

Interesting; I'll have a look at #100.

PR #102, opened 2021-11-26
`ghmerge`: support stopping interactive rebase for adaptations

DDvO:
Sometimes it is helpful to stop an interactive rebase, e.g., for fixing typos.
So far, `ghmerge` ignored any 'b' or 'e' commands given by the user,
and in effect the rebase was incomplete and the merge went wrong.
This is fixed here.

### DIFF
--- a/review-tools/ghmerge
+++ b/review-tools/ghmerge
@@ -230,6 +230,10 @@ if [ "$INTERACTIVE" == "yes" ] ; then
     echo -n "Press Enter to interactively rebase $AUTOSQUASH on $REMOTE/$TARGET: "; read foo
     REBASING=1
     git rebase -i $AUTOSQUASH $REMOTE/$TARGET || exit 1
+    if [ -e .git/rebase-merge ] ; then  # likely, user tried 'b' or 'e' command to break or enter edit mode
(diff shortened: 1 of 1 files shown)


mspncp:
Just a spontaneous thought: instead of polling and forcing the user to spawn a separate shell, why not just terminate and implement a `—resume` (or `—continue`) option instead?

mspncp:
OTOH, on posix operating systems, the user could simply suspend `ghmerge`, fix the problem, and resume it.

paulidale:
My workflow here is to use a second window to make changes while the first is sitting at the should this be committed prompt.  Make the changes, then say yes to commit.

paulidale (approved):

DDvO:
> Just a spontaneous thought: instead of polling and forcing the user to spawn a separate shell, why not just terminate and implement a `--resume` (or `--continue`) option instead?

This would be a hassle to implement.
Note that this shell script remembers various state information in variables, which would need to be saved and restored.

t8m (approved):

DDvO:
ready to merge

t8m:
Merged. Thank you.

PR #105, opened 2022-02-21
Don't let addrev count the author as a reviewer

mattcaswell:
As per the latest policy change the author is no longer counted as a
reviewer, so we ensure addrev disallows this.

### DIFF
--- a/review-tools/gitaddrev
+++ b/review-tools/gitaddrev
@@ -118,7 +118,7 @@ my $trivial = !! grep(/^CLA:\s*Trivial\s*$/i, @commit_message);
 # let's be cautious and check both, in case someone has been registered
 # as a known identity without having a CLA in place.
 die "Commit author ",$author," has no CLA, and this is a non-trivial commit\n"
-if !$trivial && grep { $_ eq $author } (@nocla_reviewers);
+    if !$trivial && grep { $_ eq $author } @nocla_reviewers;
 
 # Now that that's cleared, remove the author from anything that could cause
 # more unnecessary errors (false positives).
--- a/review-tools/gitaddrev
+++ b/review-tools/gitaddrev
@@ -28,14 +28,16 @@ my $query = OpenSSL::Query->new();
 my @reviewers;
 my @nocla_reviewers;
 my @unknown_reviewers;
-my $skip_reviewer;
 my $otccount = 0;
+my $author = $ENV{GIT_AUTHOR_EMAIL};
+my $authorrev = $query->find_person_tag($author, 'rev');
+
 sub try_add_reviewer {
     my $id = shift;
     my $rc = undef;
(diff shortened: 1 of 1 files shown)


levitte:
Comparing with the rev tag is a bit over the top. I'm on the move, so will review more closely later

mattcaswell:
> Comparing with the rev tag is a bit over the top. I'm on the move, so will review more closely later

The idea was to ensure that the author email from the commit was in some kind of canonical form. You could possibly assume that the author email is already in the canonical format...?

mattcaswell:
For example consider the situation where the author email address is a different email address to the author rev tag address - but they are actually the same person? I can see some actual commits where this is the case.

review-tools/gitaddrev
```diff
@@ -108,41 +111,19 @@ foreach (@ARGV) {
...
-    @nocla_reviewers = grep { $_ ne $author } @nocla_reviewers;
-    @unknown_reviewers = grep { $_ ne $author } @unknown_reviewers;
-    $skip_reviewer = $author;
-}
+# Note: it really should be enough to check if $author is unknown, since
+# the databases are supposed to be consistent with each other.  However,
+# let's be cautious and check both, in case someone has been registered
+# as a known identity without having a CLA in place.
+die "Commit author ",$author," has no CLA, and this is a non-trivial commit\n"
+if !$trivial && grep { $_ eq $author } (@nocla_reviewers);
```
levitte, line 121: This line should be indented 4 spaces (it should always have been, it was sloppy coding not to).  Also, the parenthesis is unnecessary, and frankly a bit confusing...

levitte (changes requested):
A nit

levitte:
> Comparing with the rev tag is a bit over the top. I'm on the move, so will review more closely later

It was too quick a look.  Upon looking again, I see the logic

paulidale:
> For example consider the situation where the author email address is a different email address to the author rev tag address - but they are actually the same person? I can see some actual commits where this is the case.

This only matters if it is a committer who is the author.  That reduces the risk significantly.

mattcaswell:
> This only matters if it is a committer who is the author. That reduces the risk significantly.

Ironically, the "actual commits" where this occurred that I'm talking about were yours @paulidale :-)

mattcaswell:
Fixup pushed addressing feedback. Please take another look.

t-j-h (approved):

paulidale (approved):

levitte (approved):

mattcaswell:
Does the 24h timer apply in this repo? I'm not sure. Also not sure about the 2 reviewer count (although I got that anyway).

paulidale:
@mattcaswell, they are good questions.  Moot now that the time and review count are there.

mattcaswell:
Pushed.

PR #103, opened 2021-12-03
ghmerge: The default remote is on github.openssl.org

t8m:
The tools repo has been already moved to github.openssl.org.

PR #106, opened 2022-02-28
ghmerge: fix REMOTE to use the GHE repo pattern

levitte:


review-tools/ghmerge
```diff
@@ -130,7 +130,7 @@ else
 fi
 ADDREVOPTS=${ADDREVOPTS# } # chop any leading ' '
 
-[ "$REMOTE" = "" ] && REMOTE=`git remote -v | awk '/git.openssl.org.*(push)/{ print $1; }' | head -n 1` # usually this will be 'upstream'
+[ "$REMOTE" = "" ] && REMOTE=`git remote -v | awk '/github.openssl.org:openssl/openssl.*(push)/{ print $1; }' | head -n 1` # usually this will be 'upstream'
 if [ "$REMOTE" = "" ] ; then
     echo Cannot find git remote with URL including 'git.openssl.org'
```
paulidale, line 135: Probably this one too....

review-tools/ghmerge
```diff
@@ -130,7 +130,7 @@ else
 fi
 ADDREVOPTS=${ADDREVOPTS# } # chop any leading ' '
 
-[ "$REMOTE" = "" ] && REMOTE=`git remote -v | awk '/git.openssl.org.*(push)/{ print $1; }' | head -n 1` # usually this will be 'upstream'
+[ "$REMOTE" = "" ] && REMOTE=`git remote -v | awk '/github.openssl.org:openssl/openssl.*(push)/{ print $1; }' | head -n 1` # usually this will be 'upstream'
```
paulidale, line 133: Sadly the new `/` in the URL screws things up

mspncp (approved):

review-tools/ghmerge
paulidale: It needs to be `\\/`

paulidale (changes requested):
`\\/` works.  Verified using openssl/openssl#17741.

paulidale:
> I missed that error

So did I when I made the same change when I first noticed this :(

review-tools/ghmerge
t8m, line 133: Please do not add openssl/openssl as the ghmerge was usable with web and tools repos as well before. Just checking for github.openssl.org should suffice.

t8m (changes requested):

review-tools/ghmerge
mspncp: Or rather `github.openssl.org:openssl` to exclude personal forks.

t8m: > Or rather `github.openssl.org:openssl` to exclude personal forks.

Nope, that won't work as tools are in otc/tools in GHE.

mspncp: How about `github.openssl.org:(openssl|otc)`? Don‘t know whether personal forks will be used on GHE, but it will be more robust.

mspncp: (Probably backslashes are required before the regex operators, didn‘t check)

paulidale: `github.openssl.org:(openssl|o[mt]c)` would be better.

general-policies is omc not otc.

mspncp: Sorry for forgetting about the OMC repos! But I would prefer `github.openssl.org:(openssl|omc|otc)` for better readability. Don't overengineer the regex ;-).

paulidale: I left the leading `o` in place at least :)

mspncp: You can bet that I also already thought about this further optimization. ;-)

t8m:
There is actually an older alternative PR #103 that I've long ago forgotten about.

PR #114, opened 2022-04-04
Adjust addrev requirements depending on the repository

### LINKED ISSUE #113: addrev shouldn't require two reviewers for the web or tools repos
It was decided that the main repository needed the extra pair of eyes unconditionally.
However, it's not certain that the web and tools repos were under the same kind of consideration.

levitte:
The main repo requires two reviewers unconditionally, while the web and tools
repos do not.

Furthermore, the web repo is governed by the OMC, and should therefore have
a minimum of one OMC member rather than a minimum of one OTC member reviewing.

Fixes #113


paulidale (dismissed):

review-tools/gitaddrev
```diff
@@ -107,8 +112,13 @@ foreach (@ARGV) {
...
     } elsif (/^--web$/) {
         $WHAT = 'web';
+        $min_reviewers = 1;
+        # openssl/web is governed by OMC
+        $min_otc = 0;
+        $min_omc = 1;
     } elsif (/--tools$/) {
-        $WHAT = 'tools'
+        $WHAT = 'tools';
+        $min_reviewers = 1;
```
levitte, line 121: We could do without this line, if the OTC feels that changes to the tools need two reviewers

t8m (approved):

review-tools/gitaddrev
paulidale: Personally, I'm okay with one reviewer.
We use these tools, we'll find out soon enough if they break :grinning:

mattcaswell (changes requested):
This change doesn't seem quite right to me. We still need 2 reviewers for web and tools. The difference is that the author counts as a reviewer in those repos. This change doesn't correctly handle that. Also, for web, only OMC members can be reviewers since they are the only people allowed to commit to the repository. For tools only OTC members can review since they are the committers for that repo.

levitte:
Following @mattcaswell's remark, I reworked this PR.  I will change the description accordingly

review-tools/gitaddrev
```diff
@@ -103,18 +123,25 @@ foreach (@ARGV) {
...
-	try_add_reviewer($1);
+        push @collect_possible_reviewers, $1;
     } elsif (/^--verbose$/) {
 	$verbose = 1;
     } elsif (/^--web$/) {
         $WHAT = 'web';
+        $min_authors = 1;
+        # openssl/web is governed by OMC
+        $min_otc = 0;
+        $min_omc = 1;
```
mattcaswell, line 134: For web both reviewers must be OMC (one of which may be the author) - at least I think that's the rules?? So shouldn't this be `$min_omc = 2`?

review-tools/gitaddrev
```diff
@@ -103,18 +123,25 @@ foreach (@ARGV) {
...
     } elsif (/--tools$/) {
-        $WHAT = 'tools'
+        $WHAT = 'tools';
+        $min_authors = 1;
```
mattcaswell, line 137: So in the tools repo we allow committers to be a reviewer? That probably makes sense, but I'm not sure if that's been made explicit anywhere?

mattcaswell (changes requested):

review-tools/gitaddrev
levitte: I could set `$min_otc = 2` for this one...

review-tools/gitaddrev
levitte: Ah, correct, the OMC is expected to be the authorized reviewers.
(if someone else wants to review from a purely technical standpoint, that's ok too, of course)

I'll fix it as you suggest

review-tools/gitaddrev
levitte: Just like I did for https://github.com/openssl/tools/pull/114#discussion_r865748628, I went looking for policies regarding this, and indeed, OTC are the only authorized reviewers, which makes `$min_otc = 2` the proper fix

review-tools/gitaddrev
levitte: Fixed

review-tools/gitaddrev
levitte: Fixed

review-tools/gitaddrev
```diff
@@ -29,10 +29,22 @@ my $query = OpenSSL::Query->new();
 my @reviewers;
 my @nocla_reviewers;
 my @unknown_reviewers;
+my $min_reviewers = 2;          # All source default
+my $authorcount = 0;
+my $min_authors = 0;            # Main source default
```
t8m, line 34: The name of this variable is slightly confusing as it is more like `count_authors` or `author_may_review`.

review-tools/gitaddrev
```diff
@@ -103,18 +123,28 @@ foreach (@ARGV) {
...
         $WHAT = 'web';
+        $min_authors = 1;
+        # openssl/web is governed by OMC
+        $min_otc = 0;
+        $min_omc = 2;
     } elsif (/--tools$/) {
-        $WHAT = 'tools'
+        $WHAT = 'tools';
+        $min_authors = 1;
+        # openssl/web is governed by OTC
```
t8m, line 138: Copy&paste error here: openssl/tools

review-tools/gitaddrev
t8m: Yes, a comment would be fine.

levitte: Done

review-tools/gitaddrev
levitte: Fixed

t8m (approved):

levitte:
Merged

d679fe04d09fdfeedd78717d63bb382f2ede0f22 Depending on the repo, authors may still count as a reviewer
214a1c80bfa58ec99ba6f268c7f9deeecb5438f7 The web repo is governed by the OMC...

PR #115, opened 2022-04-22
Support release review requirement bypass for 1.1.1/1.0.2

hlandau:


levitte (approved):

hlandau:
Updated to add support for gitaddrev.

release-tools/do-copyright-year
```diff
@@ -54,5 +54,5 @@ git diff-tree -r --name-status `git rev-list -1 --before=$NYD HEAD`..HEAD | \
     )
 echo Files changed: $(git status --porcelain | grep '^ *M' | wc -l)
 echo Committing change locally.
-git commit -m 'Update copyright year'
+git commit -m $'Update copyright year\n\nRelease: yes'
```
levitte, line 57: Not sure that this is necessary at this point, since trailers are added with `addrev` on approval anyway...

levitte:
May I suggest adding this?

``` diff
diff --git a/HOWTO-make-a-release.md b/HOWTO-make-a-release.md
index d378a4b..28a1219 100644
--- a/HOWTO-make-a-release.md
+++ b/HOWTO-make-a-release.md
@@ -169,8 +169,8 @@ openssl-tools.git checkout directory):
 
     $TOOLS/release-tools/do-copyright-year
 
-Obtain approval for these commits from the reviewer and add the reviewed-by
-headers as required.
+Obtain approval for these commits from the reviewer, add the release
+trailer, and add the reviewed-by trailers as required.
 
 *Do* send the auto-generated commits to the reviewer and await their
 approval.
```

levitte:
(note: "trailer" is git nomenclature, which I think we should follow to avoid confusion)

release-tools/do-copyright-year
hlandau: I guess it isn't, though it can't hurt. Removing it would assume addrev is always used after using the release scripts. I guess we don't need to assume that.

hlandau:
Updated.

t8m (approved):

mattcaswell:
Pushed.

PR #121, opened 2022-06-13
Modify purge-one-hour to take two arguments, BASEDIR and BASEURL

levitte:
Use cases, based on what we do with automation today:

For the main web: purge-one-hour /var/www/openssl https://www.openssl.org
For the docs: purge-one-hour /var/www/docs https://www.openssl.org/docs
For the blog: purge-one-hour /var/www/blog https://www.openssl.org/blog

### DIFF
--- a/akamai-tools/purge-one-hour
+++ b/akamai-tools/purge-one-hour
@@ -12,15 +12,35 @@
 use strict;
 use warnings;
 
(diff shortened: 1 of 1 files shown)


levitte:
This was the wrong spot.  Re-created as openssl/omc-tools#6

PR #124, opened 2022-07-29
Modify mkrelease.pl to have a different announce text for premium releases

levitte:
^CT-142 Closed

### DIFF
--- a/release-tools/mkrelease.pl
+++ b/release-tools/mkrelease.pl
@@ -22,6 +22,9 @@ require "release-git.pl";
 require "release-update.pl";
 require "release-version.pl";
 
+my @public_bversions = qw( 1.1.1 );
+my @premium_bversions = qw( 1.0.2 );
+
 our $debug   = 0;
 our $verbose = 0;
 my @reviewers;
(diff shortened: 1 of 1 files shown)


release-tools/mkrelease.pl
```diff
@@ -385,6 +396,51 @@ sub print_branch_info {
...
+
+   The OpenSSL project team is pleased to announce the release of
+   version $avers of our open source toolkit for SSL/TLS.
+
+   OpenSSL $avers is available for download via SFTP from the
+   following location:
+
+     * Server: ftp.openssl.org
+
+   Accessing it requires that you have registered an SSH key with us.
```
mattcaswell, line 426: Are we ready to say this yet? It probably needs a bit more discussion before we go down this route. If nothing else there needs to be some planning around communicating this change and supporting customers to set this up. It may take a while before everyone is transitioned to the new approach. It may be that we have both the old username/password way and the new ssh key way setup simultaneously for a while.

mattcaswell (changes requested):

release-tools/mkrelease.pl
levitte: Okie, I've changed it to require a premium password to be given if it's a premium version.

levitte:
PING!

mattcaswell (approved):

levitte:
Merged

856ba30c329684f0e102f09b94d324a1da1ac304 Modify mkrelease.pl to have a different announce text for premium releases

PR #109, opened 2022-03-14
3.0.0 is out so don't special case things as being deferred

iamamoose:

PR #127, opened 2022-09-28
3.0.0 is out so don't special case things as being deferred

levitte:
This is #109 re-created, 'cause it wasn't finished quite right, and @iamamoose
asked me.


levitte:
Merged

85126371f5705e6b3749eff73238b592c2ab91b3 3.0.0 is out so don't special case things as being deferred

iamamoose:
thanks

PR #128, opened 2022-10-11
The command to freeze the repository isn't quite right, fix it

levitte:


t8m (approved):

t8m:
Heh we have #122 which I've forgot to merge. But yours is better so I'll close that one without merging.

paulidale (approved):

levitte:
Merged

bfdac072d9b21a4c16fd3fb437a1663225cb8833 The command to freeze the repository isn't quite right, fix it

PR #139, opened 2023-03-23
Don't cross-post security advisories to oss-security

mattcaswell:
We should send any security advisory to oss-security separately and not cross-post it with our own lists.

We also change the text to say that security advisories should be sent to support-announce regardless of whether a premium release has been affected.

### DIFF
--- a/HOWTO-make-a-release.md
+++ b/HOWTO-make-a-release.md
@@ -426,15 +426,21 @@ Then copy the result to the temporary directory on dev.openssl.org:
 To finish, log in on dev.openssl.org and send the signed Security
 Advisory by email as the user that signed the advisory.
 
-For all releases, send them to the default set of public mailing lists:
+For all releases, send it to the default set of public mailing lists:
 
     REPLYTO="openssl@openssl.org" mutt -s "OpenSSL Security Advisory" \
             openssl-project openssl-users openssl-announce \
+            </tmp/secadv_FILENAME.txt.asc
+
+We also send it separately to oss-security (to avoid cross-posting with our
+own lists):
+
+    REPLYTO="openssl@openssl.org" mutt -s "OpenSSL Security Advisory" \
             oss-security@lists.openwall.com \
             </tmp/secadv_FILENAME.txt.asc
 
-For premium releases, send them to support-announce as well *and
-separately*:
+Finally we also, send it to support-announce as well *and separately*. We always
+do this, even if a premium release has not been affected:
 
     REPLYTO="openssl@openssl.org" mutt -s "OpenSSL Security Advisory" \
             support-announce </tmp/secadv_FILENAME.txt.asc


HOWTO-make-a-release.md
```diff
@@ -426,15 +426,21 @@ Then copy the result to the temporary directory on dev.openssl.org:
 To finish, log in on dev.openssl.org and send the signed Security
 Advisory by email as the user that signed the advisory.
 
-For all releases, send them to the default set of public mailing lists:
+For all releases, send it to the default set of public mailing lists:
 
     REPLYTO="openssl@openssl.org" mutt -s "OpenSSL Security Advisory" \
             openssl-project openssl-users openssl-announce \
+            </tmp/secadv_FILENAME.txt.asc
```
t8m, line 433: Not sure why (@levitte might have an idea) when I was sending the advisory mail I had to use full e-mail addresses, i.e. openssl-project@openssl.org, otherwise the mail was not sent. Also I had to use EMAIL="tomas@openssl.org" otherwise tomas@dev.openssl.org would be used as the From address.

One option would be to use the `sudo -u openssl` however that causes messages (because they are gpg signed) to be displayed with errors like: `Valid signature, but sender address and signer address do not match (Tomáš Mráz <tomas@openssl.org>)` - this is what I see in Evolution when I open such message.

t8m (changes requested):

HOWTO-make-a-release.md
levitte: I did some testing, and can see why the default configuration on dev.openssl.org doesn't agree with you.  The configuration isn't *wrong* per se, but it's at cross purpose with what you want to do.
 
I also recalled that I used to do this as the `openssl` user, not myself, which is something we used to do, quite a while ago now, and I've occasionally forgotten that.  As `openssl`, this works perfectly smoothly...  and the answer lies in `~openssl/.muttrc` and `~openssl/muttrc.aliases`, which you should have a look at.
So it seems that while we switched to running this command, the instructions essentially remained written for the `openssl` user.

levitte: You could also simply copy `~matt/.muttrc`...  it's a bit simpler...

HOWTO-make-a-release.md
```diff
@@ -426,15 +426,21 @@ Then copy the result to the temporary directory on dev.openssl.org:
...
             oss-security@lists.openwall.com \
             </tmp/secadv_FILENAME.txt.asc
 
-For premium releases, send them to support-announce as well *and
-separately*:
+Finally we also, send it to support-announce as well *and separately*. We always
+do this, even if a premium release has not been affected:
 
     REPLYTO="openssl@openssl.org" mutt -s "OpenSSL Security Advisory" \
             support-announce </tmp/secadv_FILENAME.txt.asc
```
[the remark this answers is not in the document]
mattcaswell: > thinking about it perhaps we should send advisories to support-announce but then they should be send there regardless of whether premium releases are affected or not.

That's already part of the change in this PR, i.e. always send it to support-announce regardless of whether a premium release is affected.

HOWTO-make-a-release.md
mattcaswell: My .muttrc is very simple!

````
$ cat .muttrc
set hostname = openssl.org
````

Should we add this to the instructions?

levitte: I think it would be wiser to add a file `/etc/Muttrc.d/openssl.rc` with that setting, os it's set for everyone once and for all.

mattcaswell: Ok. In that case I assume that it is outside of the scope of this PR to handle this issue.

levitte: Yes, of course.
I've now added `/etc/Muttrc.d/openssl-settings.rc`, it's a straight copy of `~matt/.muttrc`.  I tested it too, and it appears to do its job perfectly.

However, that still doesn't take away the need to set the EMAIL envvar to avoid having the mail sent as `{user}@dev.openssl.org`.  It's possible that mutt has a setting to fix that too...

mattcaswell: Hmm. I just use the command as stated in the instructions and don't have the EMAIL envvar set in my environment. It all seems to work ok for me.

levitte: Ah yup, you're right, the test I did demonstrates that.  So ok, that resolves this matter then.

t8m (approved):

arapov (approved):

paulidale (approved):

t8m:
Merged. Thank you.

PR #145, opened 2023-05-19
Reorganize the release HOWTO

levitte:
This splits up HOWTO-make-a-release.md into two new documents that reflect
the fact that *staging* and *publishing* a release are really two separate
things.

This also reflects that we're working towards full automation for staging
releases.

### DIFF
--- a/HOWTO-publish-a-release.md
+++ b/HOWTO-publish-a-release.md
@@ -12,9 +12,11 @@ Releases are staged by another procedure, separate from this.
     -   [Repositories](#repositories)
     -   [SSH access](#check-your-access)
 -   [Publish the release](#publish-the-release)
-    -   [on SFTP](#on-sftp) [only public releases]
-    -   [on Github](#on-github)
     -   [Update the source repositories](#update-the-source-repositories)
+    -   [Upload release files to OpenSSL downloads](#upload-release-files-to-openssl-downloads) [only public releases]
+    -   [Upload release files to Github](#upload-release-files-to-github)
+        -   [Web method](#web-method)
+        -   [GH CLI method](#gh-cli-method)
     -   [Update the release metadata](#update-the-release-metadata)
 -   [Post-publishing tasks](#post-publishing-tasks)
     -   [Check automations](#check-automations)
@@ -78,7 +80,14 @@ that user with sudo:
 
 # Publish the release
 
-## on SFTP
+## Update the source repositories
+
+Finish up by pushing your local changes to the appropriate source repo as
+instructed by `$TOOLS/release-tools/stage-release.sh`, which was performed
+when [staging the releases](HOWTO-stage-a-release.md).  You may want to
+sanity check the pushes by inserting the `-n` (dry-run) option.
+
+## Upload release files to OpenSSL downloads
 
 *BE CAREFUL*  This section makes everything visible and is therefore largely
 irreversible.  If you are performing a dry run then DO NOT perform any steps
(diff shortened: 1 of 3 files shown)


HOWTO-stage-a-release.md
```diff
@@ -0,0 +1,167 @@
...
+
+You must have OpenSSL's team key:
+
+    $ gpg --list-secret-key EFC0A467D613CB83C7ED6D30D894E2CE8B3D79F5
+    sec   rsa4096 2014-10-04 [SC] [expires: 2027-03-09]
+          EFC0A467D613CB83C7ED6D30D894E2CE8B3D79F5
+    uid           [ultimate] OpenSSL Security <openssl-security@openssl.org>
+    uid           [ultimate] OpenSSL security team <openssl-security@openssl.org>
+    uid           [ultimate] OpenSSL OMC <openssl-omc@openssl.org>
+    ssb   rsa4096 2014-10-04 [E] [expires: 2024-01-30]
```
t8m, line 87: This key needs to be added to fingerprints.txt and the user IDs properly uploaded to keys.openpgp.org key server.

It cannot be used with the next release yet because it needs to be in fingerprints.txt in an existing release first.

HOWTO-stage-a-release.md
```diff
@@ -0,0 +1,167 @@
...
+
+-   To get a quick usage reminder:
+
+        $TOOLS/release-tools/stage-release.sh --help
+
+-   To get a man-page:
+
+        $TOOLS/release-tools/stage-release.sh --manual
+
+It is generally called likt this:
```
t8m, line 145: Typo: like

HOWTO-publish-a-release.md
```diff
@@ -0,0 +1,301 @@
...
+future reference.
+
+For public releases (including security releases), verify that the tarballs
+are available via FTP:
+
+    ls /srv/ftp/source
+
+For premium releases, verify that the tarballs are available via SFTP:
+
+    ls /srv/premium
```
t8m, line 121: We do not publish premium releases on SFTP anymore.

HOWTO-publish-a-release.md
```diff
@@ -0,0 +1,301 @@
...
+
+## on Github
```
t8m, line 123: This step needs to be done AFTER the `Update source repositories` step otherwise the tags won't be present.

HOWTO-publish-a-release.md
```diff
@@ -0,0 +1,301 @@
...
+You can also look at the result at <https://www-origin.openssl.org>.
+
+## Check the website
+
+Verify that the release notes, which are built from the CHANGES.md file
+in the release, have been updated.  This is done automatically by OpenSSL
+automation; if you see a problem, check if the web build job has been
+performed yet, you may have to wait a few minutes before it kicks in.
+
+Wait for a while for the Akamai flush to work (normally within a few minutes).
```
t8m, line 203: CDN flush?

t8m (changes requested):

HOWTO-stage-a-release.md
levitte: > ... and the user IDs properly uploaded to keys.openpgp.org key server.

That part is actually only true if the key is to be searchable by user ID on keys.openpgp.org.  Of course, I'm not against that, but the actual truth is that the key is perfectly usable without them.

People who have signed the key should probably make sure to verify their addresses, though, as the signatures are otherwise very anonymous...

levitte: For the rest, yeah you are correct...  unfortunately, 'cause I distinctly remember that we talked about use the OpenSSL team key going forward, during more than one meeting.  But maybe I'm ahead of time...

HOWTO-publish-a-release.md
levitte: Thanks for the reminder

HOWTO-publish-a-release.md
levitte: I'd like to argue that they should *both* be done after the source update step.

HOWTO-stage-a-release.md
levitte: This is also very sad, 'cause it means that we will have to sign the tarballs and announcement texts when *publishing* them.  That can of course be done, but we've basically moved churn that we thought we could automate.  Ah well...

levitte: The more I think about this, the more I realize that it becomes difficult to sign the tarballs when staging them, when staging and publishing become separate tasks (which they do as soon as we start using automated staging and only do the publishing part manually).

The simplest is probably to let this PR wait until we have the team key firmly in place in `doc/fingerprints.txt` in all current release branches.  In other words, let's rush that one in!

paulidale (approved):

t8m (approved):

t8m:
As we are using this guide already, we should merge it. If there are any further changes needed, they can be in a separate PR.

@levitte if you agree, please merge.

mattcaswell (approved):

t8m:
I've merged this. Thank you @levitte.

PR #146, opened 2023-05-22
Create some performance testing tools

mattcaswell:
This PR creates some baseline performance testing tools and some helper code to run those performance tests. Some of the code is taken from mainline OpenSSL - but is copied here so that we can have a consistent set of code no matter which version of OpenSSL is being targetted for testing, i.e. so that we know the test is the same and the number for 1.1.1 can reliably be compared against the numbers for 3.1.

It is intended that this will be extended with additional scripts for more types of tests over time.


paulidale:
Looks good.

levitte (approved):
I haven't done a deep analysis the code itself, just looked at how it's layed out.  LGTM

mattcaswell:
Pushed. Thanks.

PR #150, opened 2023-05-26
Add a performance test for X509_CTX_CTX_get1_issuer()

mattcaswell:
This test just call X509_STORE_CTX_get1_issuer() in a loop with an empty
store. This function is usually called as part of chain building during a
certificate verify operation.

This PR is built on the commits in #146

### DIFF
--- a/perf/README
+++ b/perf/README
@@ -88,3 +88,13 @@ functions in blocks of 100 calls, and 100 blocks per thread, using a 512 bit RSA
 key. The number of threads to use is provided as an argument and the test
 reports the average time take to execute a block of 100
 EVP_PKEY_sign_init()/EVP_PKEY_sign() calls.
+
+x509storeissuer
+---------------
+
+Runs the function call X509_STORE_CTX_get1_issuer() repeatedly in a loop (which
+is used in certificate chain building as part of a verify operation). The test
+assumes that the default certificates directly exists but is empty. For a
+default configuration this is "/usr/local/ssl/certs". The test takes the number
+of threads to use as an argument and the test reports the average time take to
+execute a block of 100 X509_STORE_CTX_get1_issuer() calls.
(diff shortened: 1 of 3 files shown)


perf/README
```diff
@@ -60,3 +60,13 @@ threadcount - The number of threads to perform handshakes on in the test
...
+
+x509storeissuer
+---------------
+
+Runs the function call X509_STORE_CTX_get1_issuer() repeatedly in a loop (which
+is used in certificate chain building as part of a verify operation). The test
+assumes that the default certificates directly exists but is empty. For a
+default configuration this is "/usr/local/ssl/certs". The test takes the number
+of threads to use as an argument and the test reports the average time take to
+execute a block of 100 X509_STORE_CTX_get1_issuer() calls.
```
paulidale, line 72: Strictly, this isn't what it's doing.  It running everything then dividing the total time by number of blocks.  It should be running each block separately and averaging the times for each block.

Arithmetic mean probably isn't a good approach for a rate.  See e.g. [Notes on Calculating Computer Performance](https://tnm.engin.umich.edu/wp-content/uploads/sites/353/2021/06/1995_Notes_on_calculating_computer_performance.pdf).

mattcaswell: > Strictly, this isn't what it's doing. It running everything then dividing the total time by number of blocks. It should be running each block separately and averaging the times for each block.

I've changed it to do that. Possibly we might want to circle back over some of the other tests I've already put in and make a similar change.

> Arithmetic mean probably isn't a good approach for a rate.

I'd like to defer that to some future change.

mattcaswell:
I've rebased this to resolve conflicts and added a fixup to address the feedback.

perf/README
paulidale: Sounds good.

paulidale (approved):

mattcaswell:
Pushed. Thanks.

PR #151, opened 2023-05-26
release-tools/do-copyright-year: Modify files with more care

levitte:
Files were modified by running a simple in-place sed (i.e. 'sed -i').
This turns out to update the modification time on every file, even those
that remain unmodified.

The effect is that time stamps in a source directory become unreliable,
causing configdata.pm to be "mysteriously" older than (unmodified) files it
depends on, which causes a spurious reconfiguration when running 'make'
again.

To mediate, the loop is modified to take copies of the original files, run
an in-place sed on those, and only move them back to the original files if
there were any actual modifications.  That should leave time stamps alone on
unmodified files.

### DIFF
--- a/release-tools/do-copyright-year
+++ b/release-tools/do-copyright-year
@@ -47,8 +47,22 @@ git diff-tree -r --name-status `git rev-list -1 --before=$NYD HEAD`..HEAD | \
 	    if [ -d "$FILE" ]; then continue; fi
 	    (( count++ ))
 	    spin $count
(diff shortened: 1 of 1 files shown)


mattcaswell (approved):

levitte:
Merged

e7fe5ab5f660a7b965947fc097896d540f5f8de8 release-tools/do-copyright-year: Modify files with more care

t8m (approved):

PR #152, opened 2023-05-26
Add an RSA signing performance test

mattcaswell:
Unlike "speed" this performs RSA operations in parallel using threads
(as opposed to the -multi speed option which uses "fork").

This is built on the commits in #146


perf/Makefile
```diff
@@ -12,3 +12,6 @@ randbytes:	randbytes.c libperf.a
 
 handshake: handshake.c libperf.a
 	gcc -L$(TARGET_OSSL_LIBRARY_PATH) -L. -I$(TARGET_OSSL_INCLUDE_PATH) -I. -o handshake handshake.c -lperf -lcrypto -lssl
+
+rsasign:	rsasign.c libperf.a
```
paulidale, line 16: spurious tab?

paulidale (approved):
Should the key size be one that the FIPS provider can process?
People are going to ask for FIPS specific performance figures.

mattcaswell:
Thanks for the approval. Merged. I fixed some trivial conflicts during the merge and fixed the spurious tab you commented on.

> Should the key size be one that the FIPS provider can process?

We should definitely add that (but I've accepted your approval for this PR so would defer that to the future).

PR #153, opened 2023-05-29
A few fixes for release-tools/stage-release.sh

levitte:
Correcting these issues:

- The update branch and release branch were created each on different sides
  of a 'next_release_state' call, which made them differ even though there
  shouldn't be any difference (i.e. --branch was given or implied).
- Running 'do-copyright-year' was done after configuration, and may change
  Configure, which means that the following 'make' call may reconfigure and
  intentionally fail, causing release staging failure.
- In the metadata, the 'staging_update_branch' entry was set to the original
  update branch.

### DIFF
--- a/release-tools/stage-release.sh
+++ b/release-tools/stage-release.sh
@@ -521,16 +521,16 @@ EOF
     exit 1
 fi
 
+# Update the version information.  This won't save anything anywhere, yet,
+# but does check for possible next_method errors before we do bigger work.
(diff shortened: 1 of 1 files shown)


mattcaswell (approved):

levitte:
Merged

b756785860ddd15930258205e71edcf59545768c A few fixes for release-tools/stage-release.sh

PR #158, opened 2023-06-20
Update handshake.c

zengwei2000:
Added some erroneous judgments and released the pointer after the loop ended


paulidale:
@mattcaswell was the lack of error handling in the timing loop deliberate?

mattcaswell:
> @mattcaswell was the lack of error handling in the timing loop deliberate?

Yeah, I was looking to avoid the constant checking of ret. It still gets caught at the end, although you miss out on a useful error message in that case. But this really shouldn't error out...although I suppose if you run out of memory or something then you might do.

The missing `SSL_free` call is a valid fix though.

PR #162, opened 2023-07-20
perf/pemread.c: Divide per-thread iteration count by threadcount

t8m:
Otherwise with larger number of threads it takes too long to run.

### DIFF
--- a/perf/README
+++ b/perf/README
@@ -41,16 +41,16 @@ labels.
 randbytes
 ---------
 
-The randbytes test repeated calls the RAND_bytes() function in blocks of 100
-calls, and 100 blocks per thread. The number of threads to use is provided as
-an argument and the test reports the average time take to execute a block of 100
-RAND_bytes() calls.
+The randbytes test does 10000 calls of the RAND_bytes() function divided
+evenly among multiple threads. The number of threads to use is provided as
+an argument and the test reports the average time take to execute a block of
+1000 RAND_bytes() calls.
 
 handshake
 ---------
 
-Performs a combined in-memory client and server handshake. Each thread performs
-1000 such handshakes. It takes 2 arguments:
+Performs a combined in-memory client and server handshake. In total 100000
+handshakes are performed divided evenly among each thread. It takes 2 arguments:
 
 certsdir - A directory where 2 files exist (servercert.pem and serverkey.pem) for
 the server certificate and key. The test/certs directory of the main OpenSSL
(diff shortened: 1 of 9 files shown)


Alich-OpenSSL:
$ cat /proc/cpuinfo | grep "processor" | wc -l
4

$ for I in 1 2 3 4 5 10 100 1000; do echo "Threads: 1000"; (time ./pemread 1000) 2>&1 | grep -e "real" -e "PEM"; echo; done
Threads: 1
Average time per 100 PEM_read_bio_PrivateKey() calls: 282495.080000us
real    0m28.268s

Threads: 2
Average time per 100 PEM_read_bio_PrivateKey() calls: 175171.240000us
real    0m17.524s

Threads: 3
Average time per 100 PEM_read_bio_PrivateKey() calls: 126662.720000us
real    0m12.673s

Threads: 4
Average time per 100 PEM_read_bio_PrivateKey() calls: 98515.310000us
real    0m9.858s

Threads: 5
Average time per 100 PEM_read_bio_PrivateKey() calls: 100666.760000us
real    0m10.073s

Threads: 10
Average time per 100 PEM_read_bio_PrivateKey() calls: 100225.180000us
real    0m10.029s

Threads: 100
Average time per^C00 PEM_read_bio_PrivateKey() calls: 103214.720000us
real    0m10.329s

Threads: 1000
Average time per 100 PEM_read_bio_PrivateKey() calls: 112072.360000us
real    0m11.218s

paulidale:
How long is too long?

t8m:
> How long is too long?

IMO the test should perform the same amount of work regardless of the number of threads used.

mattcaswell:
> Then we should change the others too?

If we want to make this change - then yes - we should change all the others as well.

> Or, maybe we should not divide the number by number of threads but by number of threads/number of CPU cores (rounding up the number so we do not divide by zero) so not the amount of work per run is constant but the amount of work per core is constant.

That doesn't really make sense to me. Either it should be a fixed amount of work per thread (as it is now), or it should be a fixed amount of work over all.

t8m:
> > Then we should change the others too?
> 
> If we want to make this change - then yes - we should change all the others as well.
> 
> > Or, maybe we should not divide the number by number of threads but by number of threads/number of CPU cores (rounding up the number so we do not divide by zero) so not the amount of work per run is constant but the amount of work per core is constant.
> 
> That doesn't really make sense to me. Either it should be a fixed amount of work per thread (as it is now), or it should be a fixed amount of work over all.

One rationale for that is that this keeps the running time roughly the same regardless of number of threads (bar locking and context switching inefficiencies) which makes the running time more predictable and possibly scheduling the periodic runs of the perf test easier.

Alich-OpenSSL:
I believe the tests don't need to have the same "conditions" exactly and could differ from each other. We are not building unified framework here. In my opinion, what the test need to do to be usable:
  * shows the numbers that helps to determine whether performance dropped or increased
  * stability/being repeatable
  * run for reasonable time to not block other work (SW engineer waits too long for the results, other tests wait to be executed).

So in other words I don't see a problem to have different tests running with different number of threads/memory use/runtime/... while reporting results like "this particular area: didn't change/changed in a better (or worse) way".

t8m:
@mattcaswell How should I proceed with this?

mattcaswell:
How about you just update the README and explain how this test works differently to the others?

t8m:
In the end I've decided to bite the bullet and changed all the tests to do a similar thing in regards to number of iterations per thread.

PR #171, opened 2023-11-07
NixOS Fixes

hlandau:
I've been using these for ages. Probably time to get them upstreamed.

Another fix to come for pick-to-branch when using worktrees.

### DIFF
--- a/release-tools/stage-release.sh
+++ b/release-tools/stage-release.sh
@@ -1,10 +1,12 @@
-#! /bin/bash -e
+#! /usr/bin/env bash
 # Copyright 2020-2023 The OpenSSL Project Authors. All Rights Reserved.
 #
 # Licensed under the Apache License 2.0 (the "License").  You may not use
 # this file except in compliance with the License.  You can obtain a copy
 # in the file LICENSE in the source distribution or at
 # https://www.openssl.org/source/license.html
+
+set -e
 
 # This is the most shell agnostic way to specify that POSIX rules.
 POSIXLY_CORRECT=1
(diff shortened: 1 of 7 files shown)


levitte (approved):

release-tools/stage-release.sh
```diff
@@ -1,11 +1,13 @@
...
+#! /usr/bin/env bash
 # Copyright 2020-2023 The OpenSSL Project Authors. All Rights Reserved.
 #
 # Licensed under the Apache License 2.0 (the "License").  You may not use
 # this file except in compliance with the License.  You can obtain a copy
 # in the file LICENSE in the source distribution or at
 # https://www.openssl.org/source/license.html
 
+set -e
+
```
levitte, line 10: Side note: the move of `-e` was a bit surprising...  `/usr/bin/env bash -e` is as perfectly valid, and I find it more visible.  But, I'm ok with this, still approving this PR.

hlandau: Is it? UNIX shebangs are weird... some UNIX platforms allow an arbitrary number of arguments on the shebang line, but many (including Linux) are weird and make everything a single argument, basically only splitting on the first space. So `#!/bin/foo alpha beta gamma` is equivalent to `/bin/foo "alpha beta gamma"`.

mspncp: FWIW: personally, I prefer the more verbose line

```
set -o errexit
```

mspncp: Don't know whether that's a bashism though.

paulidale: The original and the best: `set -e` please.

mspncp: ... says the former participant of the IOCCC 😀

t8m:
Merged. Thank you.

PR #175, opened 2024-01-17
Allow for setting of extra cflags

nhorman:
When I build to run these tests under perf, I need to include a few extra cflags (notably -rpath an -fno-omit-frame-pointer).  They're not generally needed, but it would be nice to be able to set them easily in an environment.

Add CFLAGS and LDFLAGS support to allow for easy configuration


t8m:
Should we just support CFLAGS and LDFLAGS?

nhorman:
We could do that instead, sure.

t8m (approved):

hlandau (approved):

t8m:
Merged. Thank you.

PR #176, opened 2024-01-18
update ghmerge to python3

nhorman:
using ghmerge currently creates errors for me, in the form of inability to load various python packages, as fedora moved to python3 several releases ago

propose updating the use of python to python3 here to support recent distributions.

### DIFF
--- a/review-tools/ghmerge
+++ b/review-tools/ghmerge
@@ -152,7 +152,7 @@ if ! wget --quiet $PR_URL -O /tmp/gh$$; then
     echo "Error getting $PR_URL"
     exit 1
 fi
-set -- `python -c '
+set -- `python3 -c '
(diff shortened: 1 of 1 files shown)


hlandau (approved):

t8m:
`python` executes Python3 on my Fedora installation so I never needed this. Not sure what's different on yours installation.

nhorman:
@t8m thats an interesting question.  the python binary doesn't exist on my Fedora 39 system and requires that python2/3 be explicitly specified.  Somehow the ghmerge script is mapping python to python2

t8m (approved):
As I assume python3 is now available everywhere, this should be harmless, so approving.

t8m:
Merged. Thank you.

PR #181, opened 2024-02-19
correct measurement on handshake test

nhorman:
Noticed recently that the stats for the handshake test seemed wrong

Running the following command:
time ./handshake ~/git/openssl/test/certs 1
Average time per handshake: 1319us
Handshakes per second: 758.038638

real	2m11.928s
user	2m10.708s
sys	0m0.246s

Knowing that the handshake test preforms a fixed number of handshakes per run (100000), we should be able to say that an average time of 1319us per handshake should give us the approximate run time of the test: (1319 * 100000) / 10^6 (us/s) = 131 seconds
and 758 * 131 sec = ~100000 handshakes

So for a single thread, this looks good.  But as we increase the thread count: time ./handshake ~/git/openssl/test/certs 4
Average time per handshake: 1385us
Handshakes per second: 2842.024101

real	0m35.195s
user	2m17.285s
sys	0m0.267s

(1385 us/hs * 100000 hs) / (10^6 us/s) = 138.5 seconds which doesn't match the overall runtime of the test at all

What appears to be happening is an aliasing of the run time of each thread.  In the handshake test we indpendently record the execution time of each thread, and sum them to obtain a total execution time instead of just using the overall duration returned by the perflib function that executes the test.  This suffers from preemption aliasing.  i.e. if perflib creates a thread, and that thread pre-empts the main thread before it spawn any subsequent threads, those subsequent threads account for the run time accumulated by the first thread, leading to inaccurate measurements overall.

It would be better to record the entire duration of time all threads take to complete, as produced by the duration value returned from perflib.  doing so with this patch leads to consistent results vis a vis the total test execution time:
time ./handshake ~/git/openssl/test/certs 4
Average time per handshake: 377us
Handshakes per second: 2650.504386

real	0m37.737s
user	2m27.042s
sys	0m0.308s

(377 us/hs * 100000 hs) = 37.7 sec (correct vs time output) 2650.504 handshakes/s * 37.7 sec = ~100000 handshakes (correct vs hard coded value)

### DIFF
--- a/perf/handshake.c
+++ b/perf/handshake.c
@@ -19,8 +19,6 @@ int err = 0;
 
 static SSL_CTX *sctx = NULL, *cctx = NULL;
 
-OSSL_TIME *times;
-
 static int threadcount;
 
 static void do_handshake(size_t num)
@@ -28,9 +26,6 @@ static void do_handshake(size_t num)
     SSL *clientssl = NULL, *serverssl = NULL;
     int ret = 1;
     int i;
-    OSSL_TIME start, end;
-
-    start = ossl_time_now();
 
     for (i = 0; i < NUM_HANDSHAKES_PER_RUN / threadcount; i++) {
         ret = perflib_create_ssl_objects(sctx, cctx, &serverssl, &clientssl,
@@ -40,9 +35,6 @@ static void do_handshake(size_t num)
         perflib_shutdown_ssl_connection(serverssl, clientssl);
         serverssl = clientssl = NULL;
     }
-
-    end = ossl_time_now();
-    times[num] = ossl_time_subtract(end, start);
 
     if (!ret)
         err = 1;
@@ -87,12 +79,6 @@ int main(int argc, char *argv[])
         goto err;
     }
 
-    times = OPENSSL_malloc(sizeof(OSSL_TIME) * threadcount);
-    if (times == NULL) {
-        printf("Failed to create times array\n");
-        goto err;
-    }
-
     if (!perflib_create_ssl_ctx_pair(TLS_server_method(), TLS_client_method(),
                                      0, 0, &sctx, &cctx, cert, privkey)) {
         printf("Failed to create SSL_CTX pair\n");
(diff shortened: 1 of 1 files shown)


t8m:
The cause is IMO different than you describe above and I am not sure your change is correct.

When running the test with multiple threads on a CPU with multiple cores there are multiple handshakes happening simultaneously. In your example with 4 threads assuming that you have a CPU with at least 4 cores if an average handshake takes 1385us you can do 4/0.001385 handshakes per second = 2888 handshakes per second which is roughly equal to 2842 as reported. So IMO the reporting is correct as is.

t8m:
When comparing the average time per handshake across different number of threads it gives you the easiest way to compare the overhead of the lock & cache contention across the threads.

nhorman:
Yes, we do multiple handshakes in parallel with multiple threads but in commit f86e03b3fd7721fc79478b9c2f6c1b5b34596dd5 you altered the test such that each thread did :
```
NUM_HANDSHAKES_PER_RUN/threadcount
```
handshakes, so no matter how many handshakes we're doing in each thread, the total is still the same

comparing the average time per handshake * NUM_HANDSHAKES_PER_RUN against the total real time taken by the test, as I did above with the test as it currently exists with 4 threads
(1385us/hs * 100000 hs) / 10^6 us/s = 138.5 s vs a total real time on the test of 35 seconds seems pretty clearly erroneous to me.

The number of handshakes per second was reported correctly, I just updated the math in the test to make it a bit more clear how we're computing it (the average number of hs/s * the real time spent in the test should always come out to about the same number (NUM_HANDSHAKES_PER_RUN).  I can revert that bit if you like, but it seemed more clear to me this way

t8m:
> time ./handshake ~/git/openssl/test/certs 4 Average time per handshake: 377us Handshakes per second: 2650.504386
> 
> real 0m37.737s user 2m27.042s sys 0m0.308s

The results for average time per handshake above with your patch are not correct - it does not take 377us to complete a single handshake. In reality it takes about 4 times longer time to complete it but thanks to 4 cores and relatively low overhead we are able to complete 4 of them in 4*377us = 1508us simultaneously.

nhorman:
Then the labeling is wrong, the average time to complete a handshake is not 1508 us, thats the average time to complete 4 handshakes in parallel.  the unit metric for what you are describing is average (cpu*handshakes)/usecond not handshakes/second

PR #182, opened 2024-02-23
fix variable aliasing in x509storeissuer

nhorman:
threadcount is defined globally and in main, which is fine, but doing so causes assignments to use the local scope rather than the global, which results in div-by-zero in the thread function when we divide NUM_CALLS by threadcount

### DIFF
--- a/perf/x509storeissuer.c
+++ b/perf/x509storeissuer.c
@@ -65,7 +65,7 @@ static void do_x509storeissuer(size_t num)
 
 int main(int argc, char *argv[])
 {
-    int threadcount, i;
+    int i;
     OSSL_TIME duration, av;
     uint64_t us;
     double avcalltime;


t8m (approved):

hlandau (approved):

nhorman:
I'm not sure how I could have, I expressly had to add a remote for the repo to do this merge as I had the ask @t8m what the process was

mattcaswell:
Nonetheless it does not appear to be in the commit list AFAICT:

https://github.com/openssl/tools/commits/master/

Yet it does appear in @Sashan's clone:

https://github.com/sashan/tools/commits/master/

Which suggests it was once in this github.com version of the repo - but is no longer there. It does not exist in the GHE version that this repo is mirrored from

nhorman:
roger that, no idea how this happened, but let me re-merge it to the ghe instance

mattcaswell:
I've pushed it now.

nhorman:
thank you!

PR #184, opened 2024-03-09
Fix stats

nhorman:
The stats measurements in the tools repo are inconsistent.  In some tests (like handshake) we measure each threads run time, add them and average them across the total numer of measured calls, while in others we just take the aggregate run time of the test.  And in many we average that by a random value that was at some time in the past meaningful (usually NUM_CALLS_PER_BLOCK), but is no longer useful, as most tests divide the number of calls to make across the number of threads being used.  Bring them all into line using a consistent measurement scheme:
1) For a given test, divide the number of calls by the threadcount
2) record the run time for each thread
3) after the threads complete, add each thread run time
4) divide the total aggregate thread execution time by the number of calls made

Also, because some tests are fast, don't use ossl_time2us() as that uses integer division to translate a number of ticks to us, which results loss of data when measurements take less than an average of 1us.  Instead use ossl_time2ticks, casting to a double to get the proper fractional value

### DIFF
--- a/perf/newrawkey.c
+++ b/perf/newrawkey.c
@@ -111,7 +111,7 @@ int main(int argc, char *argv[])
     if (terse)
         printf("%lf\n", av);
     else
-        printf("Average time per EVP_PKEY_new_raw_public_key_ex() calls: %lfus\n",
+        printf("Average time per EVP_PKEY_new_raw_public_key_ex() call: %lfus\n",
                av);
 
     rc = EXIT_SUCCESS;
--- a/perf/newrawkey.c
+++ b/perf/newrawkey.c
@@ -13,10 +13,9 @@
 #include <openssl/evp.h>
 #include "perflib/perflib.h"
 
-#define NUM_CALLS_PER_BLOCK         1000
-#define NUM_CALL_BLOCKS_PER_RUN     100
-#define NUM_CALLS_PER_RUN           (NUM_CALLS_PER_BLOCK * NUM_CALL_BLOCKS_PER_RUN)
+#define NUM_CALLS_PER_TEST         100000
 
+OSSL_TIME *times;
 
 int err = 0;
 
@@ -32,8 +31,11 @@ void do_newrawkey(size_t num)
 {
     int i;
     EVP_PKEY *pkey;
+    OSSL_TIME start, end;
 
-    for (i = 0; i < NUM_CALLS_PER_RUN / threadcount; i++) {
+    start = ossl_time_now();
+
+    for (i = 0; i < NUM_CALLS_PER_TEST / threadcount; i++) {
         pkey = EVP_PKEY_new_raw_public_key_ex(NULL, "X25519", NULL, buf,
                                               sizeof(buf));
         if (pkey == NULL)
@@ -41,15 +43,20 @@ void do_newrawkey(size_t num)
         else
             EVP_PKEY_free(pkey);
     }
+
+    end = ossl_time_now();
+    times[num] = ossl_time_subtract(end, start);
 }
 
 int main(int argc, char *argv[])
 {
     OSSL_TIME duration;
-    uint64_t us;
-    double avcalltime;
+    OSSL_TIME us;
+    double av;
     int terse = 0;
     int argnext;
+    size_t i;
+    int rc = EXIT_FAILURE;
 
     if ((argc != 2 && argc != 3)
                 || (argc == 3 && strcmp("--terse", argv[1]) != 0)) {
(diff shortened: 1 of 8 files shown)


perf/newrawkey.c
```diff
@@ -70,25 +77,45 @@ int main(int argc, char *argv[])
...
+     */
+    av = (double)ossl_time2ticks(us)/(double)OSSL_TIME_US;
 
     if (terse)
-        printf("%lf\n", avcalltime);
+        printf("%lf\n", av);
     else
-        printf("Average time per %d EVP_PKEY_new_raw_public_key_ex() calls: %lfus\n",
-               NUM_CALLS_PER_BLOCK, avcalltime);
+        printf("Average time per EVP_PKEY_new_raw_public_key_ex() calls: %lfus\n",
```
t8m, line 114: Are these actually now `Average time per one EVP_PKEY_new_raw_public_key_ex() call:` and similarly below?

t8m (changes requested):

perf/newrawkey.c
nhorman: Yes, thats my interpretation of how this works now:
1) Each spawned thread marks the start  time of their loop
2) Each thread calls EVP_PKEY_new_raw_public_key_ex NUM_CALLS_PER_TEST/threadcount times
3) Each spawned thread marks the end time of their loop, subtracts the start time from it, and records it
4) When all threads are joined, the main routine adds each threads duration (from (2)), and divides that aggregate  value by NUM_CALLS_PER_TEST

As such SUM(i=0..threadcount, thread_duration[i]) / NUM_CALLS_PER_TEST should be the average call time for a single call to EVP_PKEY_new_raw_public_key

t8m: So, could you please change the text message as the plural there does not make sense to me anymore.

nhorman: handled, i audited the tests and removed the plural from tests, as they all measure the average single call time now

t8m: You didn't seem to push the commits?

t8m (approved):

perf/newrawkey.c
```diff
@@ -70,25 +77,45 @@ int main(int argc, char *argv[])
...
+        goto out;
     }
 
-    us = ossl_time2us(duration);
-
-    avcalltime = (double)us / NUM_CALL_BLOCKS_PER_RUN;
+    us = times[0];
+    for (i = 1; i < threadcount; i++)
+        us = ossl_time_add(us, times[i]);
+    us = ossl_time_divide(us, NUM_CALLS_PER_TEST);
```
mattcaswell, line 99: Shouldn't we convert `us` to double before this divide?

mattcaswell (changes requested):

perf/newrawkey.c
nhorman: I do that below when converting to an average number of microseconds.  I suppose we could do it here in all the tests instead.  i.e.:
```
av = ((double)(ossl_time_to_ticks(us)/(double)(NUM_CALLS_PER_TEST)) / (double)(OSSL_TIME_US)
```
Do you think that would make more sense?

Sashan (approved):

t8m:
Needs a rebase

perf/newrawkey.c
```diff
@@ -32,24 +31,32 @@ void do_newrawkey(size_t num)
...
+    end = ossl_time_now();
+    times[num] = ossl_time_subtract(end, start);
 }
 
 int main(int argc, char *argv[])
 {
     OSSL_TIME duration;
-    uint64_t us;
-    double avcalltime;
+    OSSL_TIME us;
```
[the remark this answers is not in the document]
nhorman: I think I had that fixed up in the master branch while working on this.  rebase has taken care of it

t8m (approved):

t8m:
ping @mattcaswell for re-review

perf/newrawkey.c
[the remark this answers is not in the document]
mattcaswell: Has it? It still seems to be called `us`

perf/newrawkey.c
mattcaswell: This also doesn't seem to have been changed...did you push your changes?

mattcaswell (changes requested):

nhorman:
@mattcaswell I did push my changes, but you asked for the us renaming in one file, and I didn't think to check the others, will fix

nhorman:
fixed

t8m (approved):

perf/newrawkey.c
```diff
@@ -70,25 +77,44 @@ int main(int argc, char *argv[])
...
+        ttime = ossl_time_add(ttime, times[i]);
 
-    avcalltime = (double)us / NUM_CALL_BLOCKS_PER_RUN;
+    /*
+     * EVP_PKEY_new_raw_public_key is pretty fast, running in
+     * only a few us.  But ossl_time2us does integer division
+     * and so because the average us computed above is less than
+     * the value of OSSL_TIME_US, we wind up with truncation to
+     * zero in the math.  Instead, manually do the division, casting
+     * our values as doubles so that we comput the proper time
```
mattcaswell, line 106: typo: compute

perf/randbytes.c
```diff
@@ -59,25 +67,37 @@ int main(int argc, char *argv[])
...
         printf("Error during test\n");
-        return EXIT_FAILURE;
+        goto out;
     }
 
-    us = ossl_time2us(duration);
+    us = times[0];
+    for (i = 1; i < threadcount; i++)
+        us = ossl_time_add(us, times[i]);
+    us = ossl_time_divide(us, NUM_CALLS_PER_TEST);
```
mattcaswell, line 89: Was this one missed?

mattcaswell (changes requested):

mattcaswell (approved):

mattcaswell:
@t8m please reconfirm

t8m (approved):

nhorman:
fixed, new system didn't have my config setup properly

t8m:
Merged. Thank you.

PR #188, opened 2024-04-11
Adjust Makefile for non-gnu OSes

Sashan:
Let Makefile to use CC variable to chose compiler conveniently. Rename current CFLAGS to CPPFLAGS.

Also we need to pass '-pthread' option on OpenBSD. The option is carried in CFLAGS now.

### DIFF
--- a/perf/Makefile
+++ b/perf/Makefile
@@ -5,33 +5,34 @@ clean:
 
 #-Wl,-rpath,$(TARGET_OSSL_LIBRARY_PATH)
 
-CFLAGS += -I$(TARGET_OSSL_INCLUDE_PATH) -I.
+CPPFLAGS += -I$(TARGET_OSSL_INCLUDE_PATH) -I.
+CFLAGS += -pthread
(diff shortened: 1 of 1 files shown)


perf/Makefile
```diff
@@ -5,33 +5,34 @@ clean:
 
 #-Wl,-rpath,$(TARGET_OSSL_LIBRARY_PATH)
 
-CFLAGS += -I$(TARGET_OSSL_INCLUDE_PATH) -I.
+CPPFLAGS += -I$(TARGET_OSSL_INCLUDE_PATH) -I.
+CFLAGS += -lpthread
```
levitte, line 9: Strictly speaking, `-lpthread` should be part of `LDFLAGS`, not `CFLAGS`

Sashan: Thanks for cathing this. `-lpthread` is typo. I meant to write `-pthread`. I will re-test updated Makefile and push updated change.

levitte (approved):

nhorman (approved):

levitte:
This needs a second @openssl/otc approval

levitte:
Merged

PR #190, opened 2024-04-12
Add label validation to prs

nhorman:
Add a check to ensure that we don't merge prs that don't have severity: urgent or approval: ready to merge set

### DIFF
--- a/review-tools/ghmerge
+++ b/review-tools/ghmerge
@@ -147,8 +147,9 @@ if [ "$PRNUM" = "" -o "$TEAM" = "" ] ; then
     usage_exit
 fi
 
+PR_URL_CONTENTS=$(mktemp /tmp/gh.XXXXXX)
 PR_URL=https://api.github.com/repos/openssl/$WHAT/pulls/$PRNUM
-if ! wget --quiet $PR_URL -O /tmp/gh$$; then
+if ! wget --quiet $PR_URL -O $PR_URL_CONTENTS; then
     echo "Error getting $PR_URL"
     exit 1
 fi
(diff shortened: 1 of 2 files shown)


review-tools/ghmerge
```diff
@@ -157,24 +158,48 @@ from __future__ import print_function
...
+        rtm_set=1
+    if (l["name"] == "severity: urgent"):
+        urgent_set=1
+if (rtm_set == 0 and urgent_set == 0):
+    print("Not ready to merge")
+' <$PR_URL_CONTENTS)
+
+if [ "$REPO_CHECK" == "Not ready to merge" ]
+then
+    >&2 echo "This pr has neither the urgen or ready to merge flag set, Won't merge"
```
t8m, line 191: Typo: urgent

t8m (approved):

levitte:
Side note: all those python scriplets to filter json output could easily be refactored into `jq` commands.  That would be a worthy modernization...

levitte (approved):

t8m:
Merged. Thank you.

nhorman:
@levitte I had actually considered doing that, but wasn't sure how widely distributed jq was on our various development systems.  if there is general consensus on doing that, I can follow up with a refactor

levitte:
I would imagine that jq is a commonly available on most Linuxes at this point.  It's been a commonly available on Debian since [buster](https://packages.debian.org/buster/jq), which goes like what, 5 years back now?  It's also been commonly available on Ubuntu since [20.04LTS](https://packages.ubuntu.com/search?keywords=jq).

levitte:
... and the availability doesn't stop there: https://jqlang.github.io/jq/download/

nhorman:
yeah, availability isn't a problem, was more concerned with the fact that its not part of the default install.  But that can be fixed with an additional check.  I'll make the change

levitte:
Additional check sounds good

PR #191, opened 2024-04-13
Benchmark for decoder performance (privkey) #454

Sashan:
pemread.c/pemread gets renamed to pkeyread.c/pkeyread
The tool can newly benchmak RSA, DSA, DH, DHX, EC, XKEY.

To run benchmark for all keys and both formats (PEM, DER)
using four threads execute pkeyread command as follows:
```
./pkeyread -k all -f all -t 4
```
The output will look like this:
```
lifty$ ./pkeyread -k all -f all -t 4
[dh pem] 317963.170000
[dh der] 323924.870000
[dhx pem] 76381.500000
[dhx der] 81370.580000
[dsa pem] 84088.790000
[dsa der] 84630.180000
[ec pem] 24644.910000
[ec der] 77426.390000
[rsa pem] 41520.410000
[rsa der] 65746.920000
[xkey pem] 17786.360000
[xkey der] 20157.880000
```
Change also adds `genkey.sh` which generates private keys
for testing. The script is called by `Makefile` as `regen_key_samples`
target. `Makefile` generates the key samples into `keys.h`.
The `keys.h` file is also committed to minimize a builttime
dependency on `genkey.sh` script.

To add a new key one should update genkey.sh script,
run `make regen_key_samples` and commit new version
of `keys.h` file.


perf/pkeyread.c
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    if (fail) {
+        printf("Error during test\n");
+        return EXIT_FAILURE;
+    }
+
+    us = ossl_time2us(duration);
+
+    avcalltime = (double)us / NUM_CALL_BLOCKS_PER_RUN;
```
nhorman: @t8m  and I debated this a while back.  While there is nothing specifically incorrect about doing the measurement this way, it leads to a behavior in which upping the thread count leads to decreasing per read times that are not particularly realistic (i.e. 100000 iterations performed by 4 threads completes in half the time that doing so with 2 threads does).  While accurate in total time, it doesn't really represent the average time per call to pkcs8_decode_batch.

Instead, the thought was to measure each threads execution time and average those numbers, see:
https://github.com/openssl/tools/pull/184

Sashan: Thanks Neil I took your changes to `pkeyread.c` The output with your changes in looks as follows:
```
lifty$ ./pkeyread -k all -f all -t 4 -v 
Average time per PEM_read_bio_PrivateKey(dh) calls: 1021.203000us
Average time per X509_PUBKEY_get0_param(dh) calls: 1057.180000us
Average time per PEM_read_bio_PrivateKey(dhx) calls: 247.702000us
Average time per X509_PUBKEY_get0_param(dhx) calls: 265.998000us
Average time per PEM_read_bio_PrivateKey(dsa) calls: 270.647000us
Average time per X509_PUBKEY_get0_param(dsa) calls: 313.164000us
Average time per PEM_read_bio_PrivateKey(ec) calls: 136.857000us
Average time per X509_PUBKEY_get0_param(ec) calls: 308.600000us
Average time per PEM_read_bio_PrivateKey(rsa) calls: 88.839000us
Average time per X509_PUBKEY_get0_param(rsa) calls: 159.471000us
Average time per PEM_read_bio_PrivateKey(xkey) calls: 98.997000us
Average time per X509_PUBKEY_get0_param(xkey) calls: 70.429000us
```

The pull request will be updated shortly.

nhorman (approved):

Sashan:
rebased on current.

perf/genkeys.sh
```diff
@@ -0,0 +1,261 @@
...
+# Licensed under the Apache License 2.0 (the "License").  You may not use
+# this file except in compliance with the License.  You can obtain a copy
+# in the file LICENSE in the source distribution or at
+# https://www.openssl.org/source/license.html
+#
+
+#
+# Script generates RSA, X25519, DSA, DH, DHX and EC private keys.
+# Keys are saved in PEM and DER formats. Script also converts
+# them keys to C-friendly definitions. Each key is saved to ASCIIZ
```
mattcaswell, line 15: s/them/the/

perf/keys.h
mattcaswell, line 1: Shouldn't this file have the standard licence blurb at the beginning?

perf/pkeyread.c
```diff
@@ -0,0 +1,381 @@
...
+
+    if (key_id == SAMPLE_ALL) {
+        if (format_id == FORMAT_ALL) {
+            /* run all samples for all formats */
+            for (k = 0; k < SAMPLE_ALL; k++) {
+                for (f = 0; f < FORMAT_ALL; f++) {
+                    sample_id = k;
+                    if (!perflib_run_multi_thread_test(do_f[f], threadcount, &duration)) {
+                        fprintf(stderr, "Failed to run the test %s in %s format]\n",
+                            sample_names[k], format_names[f]);
```
mattcaswell, line 293: Indentation seems off here

perf/pkeyread.c
```diff
@@ -0,0 +1,381 @@
...
+                sample_names[key_id], format_names[format_id]);
+            return EXIT_FAILURE;
+        }
+        if (verbose)
+            printf("Average time per %s(%s) calls: %lfus\n",
+                fname[format_id], sample_names[key_id], get_avcalltime());
+        else
+            printf("[%s %s] %lf\n", sample_names[key_id],
+                format_names[format_id], get_avcalltime());
+    }
```
mattcaswell, line 377: I disklike all the code duplication here. We have four variants of essentially the same code. Surely we can collapse all this into one variant?

perf/pkeyread.c
```diff
@@ -0,0 +1,381 @@
...
+        return;
+    }
+
+    keydata = sample_keys[sample_id][FORMAT_PEM];
+    keydata_sz = sample_key_sizes[sample_id][FORMAT_PEM];
+    pem = BIO_new_mem_buf(keydata, keydata_sz);
+
+    if (pem == NULL) {
+        fprintf(stderr, "%s Cannot create mem BIO [%s PEM]\n",
+            __func__, sample_names[sample_id]);
```
mattcaswell, line 55: Indentation seems off (I would line up `__func__` with `stderr`)

perf/pkeyread.c
```diff
@@ -0,0 +1,381 @@
...
+
+    /*
+     * Technically this includes the EVP_PKEY_free() in the timing - but I
+     * think we can live with that
+     */
+    for (i = 0; i < NUM_CALLS_PER_TEST / threadcount; i++) {
+        key = PEM_read_bio_PrivateKey(pem, NULL, NULL, NULL);
+        if (key == NULL) {
+            fprintf(stderr, "Failed to create key: %d [%s PEM]\n", i,
+                sample_names[sample_id]);
```
mattcaswell, line 70: Indentation

perf/pkeyread.c
```diff
@@ -0,0 +1,381 @@
...
+        err = 1;
+        return;
+    }
+
+    keydata = sample_keys[sample_id][FORMAT_DER];
+    keydata_sz = sample_key_sizes[sample_id][FORMAT_DER];
+    der = BIO_new_mem_buf(keydata, keydata_sz);
+    if (der == NULL) {
+        fprintf(stderr, "%s Cannot create mem BIO [%s DER]\n",
+            __func__, sample_names[sample_id]);
```
mattcaswell, line 109: Indentation (and so on for other instances like this below)

perf/pkeyread.c
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        if (X509_PUBKEY_set(&key, pkey) == 0) {
+            fprintf(stderr, "%s X509_PUBKEY_set() failed [%s BER]\n",
+                __func__, sample_names[sample_id]);
+            err = 1;
+            goto error;
+        }
+
+        if (X509_PUBKEY_get0_param(NULL, &pk_buf, &pk_buf_len, NULL, key) == 0) {
```
mattcaswell, line 140: Not really sure why we do the X509_PUBKEY_set and X509_PUBKEY_get0_param calls here. This is primarily intended as a decoder performance test. But these calls don't have much to do with decoding.

mattcaswell (changes requested):
Please also update perf/README.

It would be nice if we could generate a table showing the relative performance for various formats across versions (including 1.1.1 as a baseline)

perf/pkeyread.c
Sashan: I took them from code sample which is referenced here:
https://gist.github.com/richsalz/31a98a3095fa36ab6a66082a0c028a6f

do you want me to remove calls to `X509_PUBKEY_*()` functions?

Sashan:
New revision is rebased on current. the list of changes is as follows
   - introduced `report_result()` function to remove code duplication (pointed out by @mattcaswell)
   - fixed indentation at `fprintf()` functions (by @mattcaswell )
   - fixed typo in `genkeys.sh`
   - updated README
 
opened issues:
    - Shall I add license header to `keys.h` which is generated by `genkeys.sh`
    - Shall I remove calls to `X509_PUBKEY_*()` found in `do_derread()` function

perf/pkeyread.c
mattcaswell: Yes, I would remove them. They do not seem relevant to the main point of the test.

perf/pkeyread.c
Sashan: The duplicate code which prints result got moved to `report_result() function. Not sure what you mean here. There are four different cases we want to test:
  - single key type, single format
  - single key type, both formats
  - all key types, single format
  - all key types all formats

I don't know how to combine those four cases to limit duplication.

Sashan:
the new version will arrive shortly, resolved items:
   - add a copyright header to `keys.h`
   - remove calls to `X509_PUBKEY_*()` functions
 
item which needs further clarification:
   - remove duplicate code.

perf/pkeyread.c
```diff
@@ -0,0 +1,337 @@
...
+    } else {
+        /* run one sample in desired format */
+        sample_id = key_id;
+        if (!perflib_run_multi_thread_test(do_f[format_id], threadcount, &duration)) {
+            fprintf(stderr, "Failed to run the test %s in %s format]\n",
+                    sample_names[key_id], format_names[format_id]);
+            return EXIT_FAILURE;
+        }
+        report_result(key_id, format_id, verbose);
+    }
```
mattcaswell, line 333: There is still a lot of code duplication here. Here's one way you could reduce it (not tested):
```suggestion
    if (key_id == SAMPLE_ALL) {
        key_id_min = 0;
        key_id_max = SAMPLE_ALL;
    } else {
        key_id_min = key_id;
        key_id_max = key_id + 1;
    }
    if (format_id == FORMAT_ALL) {
        format_id_min = 0;
        format_id_max = FORMAT_ALL;
    } else {
        format_id_min = format_id;
        format_id_max = format_id + 1;
    }
    /* run samples/formats as appropriate */
    for (k = key_id_min; k < key_id_max; k++) {
        for (f = format_id_min; f < format_id_max; f++) {
            sample_id = k;
            if (!perflib_run_multi_thread_test(do_f[f], threadcount, &duration)) {
                fprintf(stderr, "Failed to run the test %s in %s format]\n",
                        sample_names[k], format_names[f]);
                return EXIT_FAILURE;
            }
            report_result(k, f, verbose);
        }
    }

```

Sashan: It works and it looks good. I like it very much thanks for the suggestion.

mattcaswell:
There seems to be an extra commit in this PR from #182. I suspect #182 was merged to the wrong repo - but your PR picked it up before it later got wiped by the next mirror.

Anyway - no big deal.

perf/pkeyread.c
```diff
@@ -0,0 +1,315 @@
...
+        if (pkey == NULL) {
+            fprintf(stderr, "%s pkey is NULL [%s BER]\n",
+                    __func__, sample_names[sample_id]);
+            err = 1;
+            goto error;
+        }
+error:
+        BIO_reset(der);
+        X509_PUBKEY_free(key);
+        key = NULL;
```
mattcaswell, line 135: `key` looks like it is no longer used and can be removed.

perf/pkeyread.c
```diff
@@ -0,0 +1,315 @@
...
+
+double get_avcalltime(void)
+{
+    int i;
+    OSSL_TIME us;
+    double avcalltime;
+
+    for (i = 1; i < threadcount; i++)
+        us = ossl_time_add(us, times[i]);
+    us = ossl_time_divide(us, NUM_CALLS_PER_TEST);
```
mattcaswell, line 182: I would do this divide as a "double" type as well.

perf/pkeyread.c
```diff
@@ -0,0 +1,315 @@
...
+        i++;
+    }
+
+    return i;
+}
+
+double get_avcalltime(void)
+{
+    int i;
+    OSSL_TIME us;
```
mattcaswell, line 177: `us` no longer seems like an appropriate name since that stands for micro-seconds - but it just represents an OSSL_TIME which isn't specifically us.

mattcaswell (changes requested):

perf/pkeyread.c
Sashan: I'm just curious: why do you prefer to use double here instead of keeping operations with integers? I took those changes from Niel, so therefore I'm asking. I somehow feel it's better to keep data as ints and convert to float in very last step. This is just my feeling.

This is the Niel's change I took that from: https://github.com/openssl/tools/pull/184

nhorman: I've since updated my PR in which this change was made, in response to a similar comment @mattcaswell made in #184 .  The long and the short of it is:
1) ossl_time_divide does integer division, and so when the divisor is 'large' we loose information in this operation.  for instance, NUM_CALLS_PER_TEST here is 1x10^5, so we could loose up to 99999 ticks in this operation, or about .1us worth of accuracy.  Casting to a double preserves that
2) the name us formerly referred to an OSSL_TIME struct which was going to be represented as microseconds, but the us variable in the new code is really just tracking a total number of ticks during the execution of the test.  And so the name ttime (total time) seemed more appropriate.

Apologies, I should have given you  a heads up about this.  But if you look at the last fixup commit on #184 you can see the changes I made there in response to matts comment

Sashan: Thank you Niel for clarification. Indeed I've missed this update in your #184 PR which fixes stats.

Sashan:
Conflicts with PR which added rwlock tests got resolved. Please check changes to Makefile and Readme file. thanks.

t8m:
Please rebase and drop the merge commits.

perf/pkeyread.c
```diff
@@ -0,0 +1,312 @@
...
+{
+    int i;
+    OSSL_TIME t;
+    double avcalltime;
+
+    for (i = 1; i < threadcount; i++)
+        t = ossl_time_add(t, times[i]);
+    avcalltime = (double)ossl_time2ticks(t) / (double)NUM_CALLS_PER_TEST;
+
+    avcalltime = (double)ossl_time2ticks(t) / (double)OSSL_TIME_US;
```
mattcaswell, line 181: This looks wrong. Shouldn't that be

```suggestion
    avcalltime = avcalltime / (double)OSSL_TIME_US;
```

mattcaswell (changes requested):

Sashan:
sorry about poor git driving, the merge commits should be gone now.

Sashan:
and now with fix to glitch spotted by matt

mattcaswell (approved):

perf/README
```diff
@@ -110,19 +110,24 @@ It does 100000 repetitions divided evenly among each thread. The number of
...
 LOCK_WRITERS environment variable).  Threads then iteratively acquire a shared
 rwlock to read or update some shared data.  The number of read and write
 lock/unlock pairs are reported as a performance measurement
+
+pkeyread
+--------
+
+The pkeyread test repeatedly calls the PEM_read_bio_PrivateKey() function on
+a memory BIO with a private RSA key, when it is running in pem mode (-f pem).
+If test is running in der mode (-f der) it calls to d2i_PKCS8_PRIV_KEY_INFO_bio()
```
t8m, line 126: IMO this is wrong. Testing d2i_PKCS8_PRIV_KEY_INFO_bio() is not that much useful. That function does not invoke decoders at all. Instead d2i_PrivateKey_ex() on the same data would provide much more useful results.

However I see that you then call EVP_PKCS82PKEY() - this is the actual function where the decoders are invoked. Funnily enough it first reverts the d2i_PKCS8_PRIV_KEY_INFO operation by calling i2d which is then passed to a decoder.

IMO to see more realistic performance of DER decoding of a PKEY d2i_PrivateKey_ex() should be tested instead of these two calls together.

t8m (changes requested):

Sashan:
the pull request will be updated shortly. The list of changes is as follows:
   - the benchmark for DER uses `d2i_PrivateKey_ex()` now. it required changes to `genkey.sh` so it also generates table of `EVP_PKEY_*` constants with keys supported by test. `keys.h` got regenrated.
   - I also made two tweaks to report results, the diff chunk for this bellow
```
@@ -168,13 +157,14 @@ static int format_name_to_id(const char *format_name)
     return i;
 }
 
-double get_avcalltime(void)
+static double get_avcalltime(void)
 {
     int i;
     OSSL_TIME t;
     double avcalltime;
 
-    for (i = 1; i < threadcount; i++)
+    memset(&t, 0, sizeof(t));
+    for (i = 0; i < threadcount; i++)
         t = ossl_time_add(t, times[i]);
     avcalltime = (double)ossl_time2ticks(t) / (double)NUM_CALLS_PER_TEST;
 
@@ -183,7 +173,7 @@ double get_avcalltime(void)
     return avcalltime;
 }
 
-void report_result(int key_id, int format_id, int verbose)
+static void report_result(int key_id, int format_id, int verbose)
 {
     if (err) {
        fprintf(stderr, "Error during test of %s in %s format\n",
@@ -192,10 +182,10 @@ void report_result(int key_id, int format_id, int verbose)
     }
 
     if (verbose)
-       printf("Average time per %s(%s) calls: %lfus\n",
+       printf("Average time per %s(%s) call: %lfus\n",
               format_names[format_id], sample_names[key_id], get_avcalltime());
     else
-       printf("[%s %s] %lf\n", sample_names[key_id],
+       printf("[%s %s] %lfus\n", sample_names[key_id],
               format_names[format_id], get_avcalltime());
 }
```

t8m:
Why it was necessary to change the key data? IMO d2i_PrivateKey_ex() should be able to decode the original data.

Sashan:
> Why it was necessary to change the key data? IMO d2i_PrivateKey_ex() should be able to decode the original data.

keys.h now contains array with EVP_PKEY_ constant. I use this array later in newly introduced function `sample_id_to_evp()`. I could add that array to keys.h by hand, but I decided to give the `genkeys.h` script a try.

t8m (approved):
@mattcaswell @nhorman please reconfirm

mattcaswell (changes requested):

Sashan:
README got updated.

Sashan:
rebased to master.

nhorman:
approval holds

t8m:
My approval holds too. Merging.

t8m:
Actually @mattcaswell could you please also reconfirm?

mattcaswell (approved):

t8m:
Merged. Thank you.

PR #192, opened 2024-04-15
Migrate ghmerge from using python to jq

nhorman:
jq is a nice handy tool to do json parsing, use it instead of python

### DIFF
--- a/review-tools/ghmerge
+++ b/review-tools/ghmerge
@@ -188,15 +188,13 @@ TARGET_REPO=$(jq -r '.base.repo.name' $PR_URL_CONTENTS)
 RTM_LABEL=$(jq -r '.labels[] | select(.name == "approval: ready to merge") | .name' $PR_URL_CONTENTS)
 URGENT_LABEL=$(jq -r '.labels[] | select(.name == "severity: urgent") | .name' $PR_URL_CONTENTS)
 
-if [ "$TARGET_REPO" == "openssl" ]
+if [ "$TARGET_REPO" != "openssl" ]
 then
-    if [ -z "$RTM_LABEL" -a -z "$URGENT_LABEL" ]
-    then
-        >&2 echo "This PR has neither the ready to merge or urgent label set, can't merge"
-        exit 1
-    fi
-else
     >&2 echo "Skipping ready to merge check for non-openssl repo"
+elif [ -z "$RTM_LABEL" -a -z "$URGENT_LABEL" ]
+then
+    >&2 echo "This PR has neither the ready to merge or urgent label set, can't merge"
+    exit 1
 fi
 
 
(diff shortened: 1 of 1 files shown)


review-tools/ghmerge
```diff
@@ -29,8 +29,19 @@ Examples:
     exit 9
 }
 
+function check_tools {
+    which jq >/dev/null 2>&1
```
levitte, line 33: I've recently learned that if you want to be more posixly, you should use [`command -v`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html) rather than `which`.

(as a matter of fact, using bash's `type` command has issued a warning on the matter lately...  that's how I learned)

t8m (approved):

review-tools/ghmerge
```diff
@@ -168,30 +176,22 @@ if [ -z "$WHO" -o -z "$BRANCH" -o -z "$REPO" ]; then
...
-if (rtm_set == 0 and urgent_set == 0):
-    print("Not ready to merge")
-' <$PR_URL_CONTENTS)
-
-if [ "$REPO_CHECK" == "Not ready to merge" ]
+TARGET_REPO=$(jq -r '.base.repo.name' $PR_URL_CONTENTS)
+RTM_LABEL=$(jq -r '.labels[] | select(.name == "approval: ready to merge") | .name' $PR_URL_CONTENTS)
+URGENT_LABEL=$(jq -r '.labels[] | select(.name == "severity: urgent") | .name' $PR_URL_CONTENTS)
+
+if [ "$TARGET_REPO" == "openssl" ]
```
Sashan, line 183: I would invert the condition here. then code can be changed like that
```
if [ "$TARGET_REPO" != "openssl" ] ;
then
   >&2 echo "Skipping ready to merge check for non-openssl repo"
elif [ -z "$RTM_LABEL" -a -z "$URGENT_LABEL" ] ;
then
  >&2 echo "This PR has neither the ready to merge or urgent label set, can't merge"
  exit 1
fi
```

Sashan (approved):
looks good

review-tools/ghmerge
nhorman: done

t8m (approved):

paulidale (approved):

paulidale (approved):
I'm not entirely convinced that this change is worthwhile.  It's not saving much in terms of lines of code, it's introducing a third party dependency and it's moving the parsing out of python into a separate process.  The benefit seems to be an improvement in a section of code that isn't oft modified and isn't that difficult to understand.

levitte (approved):

nhorman:
merged to master, thank you!

PR #193, opened 2024-05-07
adding a fetch test to perf tools

nhorman:
Add a perf test to record how many EVP fetches we can do in unit time

### DIFF
--- a/perf/evp_fetch.c
+++ b/perf/evp_fetch.c
@@ -15,7 +15,7 @@
 #include <openssl/core_names.h>
 #include "perflib/perflib.h"
 
-#define NUM_CALLS_PER_TEST         100000
+#define NUM_CALLS_PER_TEST         1000000
 
 OSSL_TIME *times;
 
@@ -261,8 +261,7 @@ int main(int argc, char *argv[])
     if (terse)
         printf("%lf\n", av);
     else
-        printf("Average time per fetch call: %lfus\n",
-               av);
+        printf("Average time per fetch call: %lfus\n", av);
 
     rc = EXIT_SUCCESS;
 out:
--- a/perf/evp_fetch.c
+++ b/perf/evp_fetch.c
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2023 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <openssl/evp.h>
+#include <openssl/kdf.h>
+#include <openssl/core_names.h>
+#include "perflib/perflib.h"
+
+#define NUM_CALLS_PER_TEST         100000
+
+OSSL_TIME *times;
+
+int err = 0;
+
+static int threadcount;
+
+static OSSL_LIB_CTX *ctx = NULL;
+
+#define ARRAY_SIZE(a)                               \
+  ((sizeof(a) / sizeof(*(a))))
+
+typedef enum {
+    FETCH_MD = 0,
+    FETCH_CIPHER,
+    FETCH_KDF,
+    FETCH_MAC,
+    FETCH_RAND
+} fetch_type_t;
+
+struct fetch_data_entry {
+    fetch_type_t ftype;
+    const char *alg;
+    const char *propq;
+};
+
+static struct fetch_data_entry fetch_entries[] = {
+    {FETCH_MD, OSSL_DIGEST_NAME_SHA2_224, NULL},
+    {FETCH_MD, OSSL_DIGEST_NAME_SHA2_256, NULL},
+    {FETCH_MD, OSSL_DIGEST_NAME_SHA3_224, NULL},
+    {FETCH_MD, OSSL_DIGEST_NAME_SHA3_256, NULL},
+    {FETCH_CIPHER, OSSL_CIPHER_NAME_AES_128_GCM_SIV, NULL},
+    {FETCH_CIPHER, OSSL_CIPHER_NAME_AES_192_GCM_SIV, NULL},
+    {FETCH_CIPHER, OSSL_CIPHER_NAME_AES_256_GCM_SIV, NULL},
+    {FETCH_KDF, OSSL_KDF_NAME_HKDF, NULL},
+    {FETCH_KDF, OSSL_KDF_NAME_SCRYPT, NULL},
+    {FETCH_KDF, OSSL_KDF_NAME_KRB5KDF, NULL},
+    {FETCH_KDF, OSSL_KDF_NAME_KBKDF, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_BLAKE2BMAC, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_CMAC, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_GMAC, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_HMAC, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_POLY1305, NULL},
+    {FETCH_RAND, "CTR-DRBG", NULL}
+};
+
+void do_fetch(size_t num)
+{
+    OSSL_TIME start, end;
+    size_t i, j;
+
+    start = ossl_time_now();
+
+    for (i = 0; i < NUM_CALLS_PER_TEST / threadcount; i++) {
+        j = i % ARRAY_SIZE(fetch_entries);
+
+        if (err == 1)
+            return;
+
+        switch (fetch_entries[j].ftype) {
+        case FETCH_MD:
+            EVP_MD *md = EVP_MD_fetch(ctx, fetch_entries[j].alg,
+                                      fetch_entries[j].propq);
+            if (md == NULL) {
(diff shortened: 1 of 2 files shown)


perf/evp_fetch.c
```diff
@@ -0,0 +1,212 @@
...
+
+    /*
+     * EVP_PKEY_new_raw_public_key is pretty fast, running in
+     * only a few us.  But ossl_time2us does integer division
+     * and so because the average us computed above is less than
+     * the value of OSSL_TIME_US, we wind up with truncation to
+     * zero in the math.  Instead, manually do the division, casting
+     * our values as doubles so that we compute the proper time
+     */
+    av = ((double)ossl_time2ticks(ttime) / (double)NUM_CALLS_PER_TEST) /(double)OSSL_TIME_US;
```
paulidale, line 199: Stylistic: only the first needs to be there, the rest will switch to double automatically.

perf/evp_fetch.c
```diff
@@ -0,0 +1,212 @@
...
+    {FETCH_KDF, OSSL_KDF_NAME_HKDF, NULL},
+    {FETCH_KDF, OSSL_KDF_NAME_SCRYPT, NULL},
+    {FETCH_KDF, OSSL_KDF_NAME_KRB5KDF, NULL},
+    {FETCH_KDF, OSSL_KDF_NAME_KBKDF, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_BLAKE2BMAC, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_CMAC, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_GMAC, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_HMAC, NULL},
+    {FETCH_MAC, OSSL_MAC_NAME_POLY1305, NULL},
+    {FETCH_RAND, "CTR-DRBG", NULL}
```
paulidale, line 62: More for the future: would storing timing information for each be useful?  I doubt there is a significant difference between each but it would be nice to know for sure.

perf/evp_fetch.c
```diff
@@ -0,0 +1,212 @@
...
+    FETCH_RAND
+} fetch_type_t;
+
+struct fetch_data_entry {
+    fetch_type_t ftype;
+    const char *alg;
+    const char *propq;
+};
+
+static struct fetch_data_entry fetch_entries[] = {
```
paulidale, line 45: The entries in this table should be conditional based on the `OPENSSL_NO_xxx` defines from `include/openssl/configuration.h`.  Alternatively, only use things that are guaranteed to exist.

perf/evp_fetch.c
```diff
@@ -0,0 +1,212 @@
...
+};
+
+void do_fetch(size_t num)
+{
+    OSSL_TIME start, end;
+    size_t i, j;
+
+    start = ossl_time_now();
+
+    for (i = 0; i < NUM_CALLS_PER_TEST / threadcount; i++) {
```
paulidale, line 72: The scaling here will introduce bias if `NUM_CALLS_PER_TEST` is not divisible by `threadcount`.  Specifically, the computation on line 199 doesn't account for any rounding from the integer division.

Suggest running the full `NUM_CALLS_PER_TEST` for each thread and compensating when averaging or change `(double)NUM_CALLS_PER_TEST` to `(double)((NUM_CALLS_PER_TEST / threadcount) * threadcount)` down line line 199.

paulidale (changes requested):
Looks good.

One subtle numeric issue.

perf/evp_fetch.c
t8m: @paulidale That would make some tests that use the same construct to run for overly long time if high number of threads is used. Is this really a problem? We can just use threadcounts that are dividing 100000 without a reminder. I.e. in the original benchmark we used 1, 10, 100, 500 and 1000 as threadcount.

nhorman: Both good points.  A happy medium might be to do the division among threads as we are currently doing, but when we do the averaging, instead of dividing by NUM_CALLS_PER_TEST, instead divide by (NUM_CALLS_PER_TEST / threadcount) * threadcount).  that would adjust the final average by the actual number of iterations performed, rather than the assumed NUM_CALLS_PER_TEST.

Would that be a reasonable approach?

perf/evp_fetch.c
nhorman: I can make this change, but I would want to do it in all the tests, as they follow this pattern.  suggest a separate pull request to do that to avoid feature creep here.

perf/evp_fetch.c
nhorman: That might be useful.  Perhaps a cmd line option or argument to direct the test to only do fetches of a specific type (or even specific alg), so that we could test something particular, or generally get a sampling of all of them.

perf/evp_fetch.c
paulidale: Yes, that's the second option I offered and it's reasonable IMO.

perf/evp_fetch.c
paulidale: No problem.  It was only a stylistic suggestion.

perf/evp_fetch.c
nhorman: updated to include this

perf/evp_fetch.c
nhorman: done

perf/evp_fetch.c
```diff
@@ -139,11 +175,13 @@ int main(int argc, char *argv[])
 {
     OSSL_TIME duration;
     OSSL_TIME ttime;
+    double real_num_calls;
     double av;
     int terse = 0;
     int argnext;
     size_t i;
     int rc = EXIT_FAILURE;
+    char *fetch_type = getenv("EVP_FETCH_TYPE");
```
paulidale, line 184: A command line opt would be easier to use...

paulidale (approved):

t8m:
Needs rebase now

nhorman:
done

paulidale (approved):

perf/evp_fetch.c
```diff
@@ -0,0 +1,272 @@
...
+        printf("Error during test\n");
+        goto out;
+    }
+
+    ttime = times[0];
+    for (i = 1; i < threadcount; i++)
+        ttime = ossl_time_add(ttime, times[i]);
+
+    /*
+     * EVP_PKEY_new_raw_public_key is pretty fast, running in
```
t8m, line 251: This comment needs correction after copy&paste.

perf/evp_fetch.c
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }
+
+    threadcount = atoi(argv[argnext]);
+    if (threadcount < 1) {
+        printf("threadcount must be > 0\n");
+        return EXIT_FAILURE;
+    }
+
+    ctx = OSSL_LIB_CTX_new();
```
t8m, line 226: I wonder if we should have a "warm up" option that would do one fetch here outside the measurements. The initial fetch on a libctx with no providers loaded will load the default provider implicitly and fighting for that could skew the numbers fairly visibly. I.e. it would be interesting to see the differences in the numbers. Or instead we could explicitly load the default provider with OSSL_PROVIDER_load().

t8m (changes requested):

perf/evp_fetch.c
nhorman: I suppose that would depend on the case we want to measure.  I could envision wanting to measure both the fast path (where a fetch would always hit in the cache), and the slow path (where a fetch would never hit in the cache)

paulidale: Another PR I think.

IMO a warm up isn't going to impact the measurement significantly given a large number of iterations per thread.

perf/evp_fetch.c
```diff
@@ -0,0 +1,272 @@
...
+
+    if (argc == 3) {
+        terse = 1;
+        argnext = 2;
+    } else {
+        argnext = 1;
+    }
+
+    if (fetch_type != NULL) {
+        exclusive_fetch_alg = strstr(fetch_type, ":");
```
Sashan, line 200: `strchr()`?  I think `strstr()` is a big hammer here.

perf/evp_fetch.c
```diff
@@ -0,0 +1,272 @@
...
+     * our values as doubles so that we compute the proper time
+     */
+    real_num_calls = ((double)NUM_CALLS_PER_TEST / threadcount) * threadcount;
+    av = ((double)ossl_time2ticks(ttime) / real_num_calls) /(double)OSSL_TIME_US;
+
+    if (terse)
+        printf("%lf\n", av);
+    else
+        printf("Average time per fetch call: %lfus\n",
+               av);
```
Sashan, line 265: No need to to break line.

Sashan (approved):
found just two nits. otherwise looks good. it's up to you if you address those nits or not. I don't insist.

paulidale (approved):

t8m (approved):

perf/evp_fetch.c
```diff
@@ -0,0 +1,272 @@
...
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <openssl/evp.h>
+#include <openssl/kdf.h>
+#include <openssl/core_names.h>
+#include "perflib/perflib.h"
+
+#define NUM_CALLS_PER_TEST         100000
```
t8m, line 18: Please consider raising this number. Even 10000000 calls takes just a few seconds on my laptop.

t8m (approved):

t8m:
Merged. Thank you for your contribution.

PR #194, opened 2024-05-17
add support for installer repo to ghmerge

nhorman:
Adds installer repo as a merge WHAT target

Note also that I did some addional gymnastics to add OpenSSL-Query to the PERL5LIB path here to avoid having to copy those modules to the installer directory

### DIFF
--- a/review-tools/gitaddrev
+++ b/review-tools/gitaddrev
@@ -148,12 +148,12 @@ foreach (@ARGV) {
         $min_omc = 0;
     } elsif (/--perftools$/) {
         $WHAT = 'perftools';
-        $min_authors = 2;
+        $min_authors = 1;
         $min_otc = 0;
         $min_omc = 0;
     } elsif (/--installer$/) {
         $WHAT = 'installer';
-        $min_authors = 2;
+        $min_authors = 1;
         # openssl/installer is governed by OTC
         $min_otc = 1;
         $min_omc = 0;
--- a/review-tools/gitaddrev
+++ b/review-tools/gitaddrev
@@ -150,9 +150,10 @@ foreach (@ARGV) {
         $WHAT = 'perftools';
         $min_authors = 2;
         $min_otc = 0;
+        $min_omc = 0;
     } elsif (/--installer$/) {
         $WHAT = 'installer';
-        $min_authors = 1;
+        $min_authors = 2;
         # openssl/installer is governed by OTC
         $min_otc = 1;
         $min_omc = 0;
--- a/review-tools/gitaddrev
+++ b/review-tools/gitaddrev
@@ -150,6 +150,11 @@ foreach (@ARGV) {
         $WHAT = 'perftools';
         $min_authors = 2;
         $min_otc = 0;
+    } elsif (/--installer$/) {
+        $WHAT = 'installer';
+        $min_authors = 1;
+        # openssl/installer is goverend by OTC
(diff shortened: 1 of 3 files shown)


t8m:
Please look at all the changes needed in this commit https://github.com/openssl/tools/commit/3ab47885e9e684c4311d5c8d138069485d087cd1

review-tools/gitaddrev
```diff
@@ -146,6 +146,12 @@ foreach (@ARGV) {
         # openssl/fuzz-corpora is governed by OTC
         $min_otc = 1;
         $min_omc = 0;
+    } elsif (/--installer$/) {
+        $WHAT = 'installer';
+        $min_authors = 1;
+        # openssl/installer is goverend by OTC
```
t8m, line 152: Typo: governed

t8m (approved):
OK with the typo corrected

t8m (approved):

review-tools/gitaddrev
```diff
@@ -150,6 +150,11 @@ foreach (@ARGV) {
         $WHAT = 'perftools';
         $min_authors = 2;
         $min_otc = 0;
+    } elsif (/--installer$/) {
+        $WHAT = 'installer';
+        $min_authors = 1;
```
mattcaswell, line 155: I might have expected this to match the main repo

review-tools/gitaddrev
```diff
@@ -150,6 +150,11 @@ foreach (@ARGV) {
...
+        # openssl/installer is governed by OTC
+        $min_otc = 1;
         $min_omc = 0;
```
mattcaswell, line 158: These are the defaults. You removed `min_omc` from `--perftools` and left `min_otc` which is strange (but it doesn't really matter since they are the default)

mattcaswell (changes requested):

review-tools/gitaddrev
nhorman: fixed this up

review-tools/gitaddrev
nhorman: made this match the other repos

review-tools/gitaddrev
```diff
@@ -151,6 +151,12 @@ foreach (@ARGV) {
         $min_authors = 2;
         $min_otc = 0;
         $min_omc = 0;
+    } elsif (/--installer$/) {
+        $WHAT = 'installer';
+        $min_authors = 2;
```
mattcaswell, line 156: A `min_authors` value of 2 is meaningless. Well actually it means the same as 1. Really the only meaningful settings are 0 and 1. 0 means the author does not count as a reviewer. 1 means the author does count. I think this variable is misnamed (it's not a minimum of anything).

I would really expect the installer repo to have the same review criteria as the main source repo (i.e. the author does not count as a reviewer).

I note `perftools` also has a `min_authors` value of 2, which is equally strange IMO.

mattcaswell (changes requested):

review-tools/gitaddrev
nhorman: fixed both to be one

t8m (approved):

t8m:
@mattcaswell please re-approve

levitte (approved):

t8m:
Merged. Thank you.

PR #195, opened 2024-05-24
Enhance handshake test to offer a variant of the test in which ssl_ctx objects are allocated per connection

nhorman:
while best practice is to create a single ssl_ctx object and share it among ssl connections, it often happens that existing code in the field creates a context per connection.  to allow us to test performance in this environment, enhance the handshake test with the -s option.  Specifying this option alters the test behavior such that contexts are created and freed per connection, rather than the default behavior which is to share a context

### DIFF
--- a/perf/README
+++ b/perf/README
@@ -52,7 +52,7 @@ handshake
 Performs a combined in-memory client and server handshake. In total 100000
 handshakes are performed divided evenly among each thread. It take 2 optional
 and two required arguments:
-handshakte [-t] [-s] <certsdir> <threadcount>
+handshake [-t] [-s] <certsdir> <threadcount>
 -t - produce terse output
 -s - create an ssl_ctx per connection, rather than a single thread-shared ctx
 certsdir - Directory where the test can locate servercert.pem and serverkey.pem


nhorman:
@mattcaswell I like the idea of a test for creating ssl_ctx objects.  I'll create a separate test for that

perf/README
```diff
@@ -50,13 +50,13 @@ handshake
...
-handshakes are performed divided evenly among each thread. It takes 2 arguments:
-
-certsdir - A directory where 2 files exist (servercert.pem and serverkey.pem) for
-the server certificate and key. The test/certs directory of the main OpenSSL
-source repository contains such files for all supported branches.
-
-threadcount - The number of threads to perform handshakes on in the test
+handshakes are performed divided evenly among each thread. It take 2 optional
+and two required arguments:
+handshakte [-t] [-s] <certsdir> <threadcount>
```
mattcaswell, line 55: typo: handshake

mattcaswell (approved):
Approved assuming the minor typo is fixed

perf/README
nhorman: fixed

paulidale (approved):

PR #196, opened 2024-06-03
Add ssl_ctx test

nhorman:
we have the handshake test, but it would be nice to have a perf test that measures just the amount of time taken to execute a SSL_CTX_new call.  Add that test here

### DIFF
--- a/perf/ssl_ctx.c
+++ b/perf/ssl_ctx.c
@@ -44,8 +44,10 @@ static void do_create_ctx(size_t num)
     for (i = 0; i < adj_ctx_per_run / threadcount; i++) {
         ctx = SSL_CTX_new(mode == OP_SERVER ? TLS_server_method() :
                                               TLS_client_method());
-        if (ctx == NULL)
+        if (ctx == NULL) {
+            err = 1;
             goto out;
+        }
         if (mode == OP_SERVER) {
             if ((SSL_CTX_use_certificate_file(ctx, certpath,
                                               SSL_FILETYPE_PEM) != 1) ||
--- a/perf/ssl_ctx.c
+++ b/perf/ssl_ctx.c
@@ -10,6 +10,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 #include <openssl/ssl.h>
 #include "perflib/perflib.h"
 
@@ -18,8 +19,18 @@
 int err = 0;
 
 OSSL_TIME *times;
+static char *certpath = NULL;
+static char *keypath = NULL;
+static char *storepath = NULL;
 
 static int threadcount;
+
+typedef enum {
+    OP_SERVER,
+    OP_CLIENT
+} op_mode;
+
+op_mode mode = OP_SERVER;
 
 static void do_create_ctx(size_t num)
 {
@@ -31,9 +42,20 @@ static void do_create_ctx(size_t num)
     start = ossl_time_now();
 
     for (i = 0; i < NUM_CTX_PER_RUN / threadcount; i++) {
-        ctx = SSL_CTX_new(TLS_server_method());
+        ctx = SSL_CTX_new(mode == OP_SERVER ? TLS_server_method() :
+                                              TLS_client_method());
         if (ctx == NULL)
             goto out;
+        if (mode == OP_SERVER) {
+            if ((SSL_CTX_use_certificate_file(ctx, certpath,
+                                              SSL_FILETYPE_PEM) == 1) ||
+                (SSL_CTX_use_PrivateKey_file(ctx, keypath,
+                                             SSL_FILETYPE_PEM) == 1))
+                goto out;
+        } else {
+            if (SSL_CTX_load_verify_dir(ctx, storepath) != 1)
+                goto out;
+        }
         SSL_CTX_free(ctx);
         ctx == NULL;
     }
@@ -46,6 +68,16 @@ out:
         err = 1;
 }
 
+static void usage(char *name)
+{
+    printf("usage\n");
+    printf("%s [-m <server|client>] [-c <cert>] [-k <key>] [-s <store>] <threadcount> \n", name);
+    printf("-m <server|client> - create a client or server method in context\n");
+    printf("-c <cert> - path to certificate for server context\n");
+    printf("-k <key> - path to key for server context\n");
+    printf("-s <store> - path to cert store for client context\n");
+}
+
 int main(int argc, char *argv[])
 {
     OSSL_TIME duration, ttime;
@@ -53,25 +85,60 @@ int main(int argc, char *argv[])
     double avcalltime;
     int ret = EXIT_FAILURE;
     int i;
-    int argnext;
     int terse = 0;
+    int ch = 0;
 
-    if ((argc == 3 && strcmp("--terse", argv[1]) != 0)) {
-        printf("Usage: ssl_ctx [--terse] threadcount\n");
-        return EXIT_FAILURE;
+    while ((ch = getopt(argc, argv, "m:c:k:s:")) != -1) {
+        switch(ch) {
+        case 'm':
+            if (!strcmp(optarg, "server")) {
+                mode = OP_SERVER;
+            } else if (!strcmp(optarg, "client")) {
+                mode = OP_CLIENT;
+            } else {
+                printf("-m must select one of client|server\n");
+                usage(argv[0]);
+            }
+            break;
+        case 'c':
+            certpath = optarg;
+            break;
+        case 'k':
+            keypath = optarg;
+            break;
+        case 's':
+            storepath = optarg;
+            break;
+        default:
+            usage(argv[0]);
+            exit(1);
+        }
     }
 
-    if (argc == 3) {
-        terse = 1;
-        argnext = 2;
-    } else {
-        argnext = 1;
+    if (argv[optind] == NULL) {
+        printf("Missing threadcount argument\n");
+        usage(argv[0]);
+        goto err;
     }
 
-    threadcount = atoi(argv[argnext]);
+    threadcount = atoi(argv[optind]);
     if (threadcount < 1) {
         printf("threadcount must be > 0\n");
         goto err;
+    }
+
+    if (mode == OP_SERVER) {
+        if (certpath == NULL | keypath == NULL) {
+            printf("server mode requires both -c and -k options\n");
+            usage(argv[0]);
+            goto err;
+        }
+    } else {
+        if (storepath == NULL) {
+            printf("client mode requires -s option\n");
+            usage(argv[0]);
+            goto err;
+        }
     }
 
     times = OPENSSL_malloc(sizeof(OSSL_TIME) * threadcount);
@@ -99,7 +166,7 @@ int main(int argc, char *argv[])
     if (terse)
         printf("%lf\n", avcalltime);
     else
-        printf("Average time per ssl_ctx call: %lfus\n", avcalltime);
+        printf("Average time per ssl ctx setup: %lfus\n", avcalltime);
 
     ret = EXIT_SUCCESS;
  err:
--- a/perf/ssl_ctx.c
+++ b/perf/ssl_ctx.c
@@ -24,6 +24,7 @@ static char *keypath = NULL;
 static char *storepath = NULL;
 
 static int threadcount;
+static int adj_ctx_per_run = 0;
 
 typedef enum {
     OP_SERVER,
@@ -41,7 +42,7 @@ static void do_create_ctx(size_t num)
 
     start = ossl_time_now();
 
-    for (i = 0; i < NUM_CTX_PER_RUN / threadcount; i++) {
+    for (i = 0; i < adj_ctx_per_run / threadcount; i++) {
         ctx = SSL_CTX_new(mode == OP_SERVER ? TLS_server_method() :
                                               TLS_client_method());
         if (ctx == NULL)
@@ -57,7 +58,6 @@ static void do_create_ctx(size_t num)
                 goto out;
         }
         SSL_CTX_free(ctx);
-        ctx == NULL;
     }
 
 out:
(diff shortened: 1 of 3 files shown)


perf/ssl_ctx.c
```diff
@@ -0,0 +1,108 @@
...
+{
+    int ret = 1;
+    int i;
+    SSL_CTX *ctx = NULL;
+    OSSL_TIME start, end;
+
+    start = ossl_time_now();
+
+    for (i = 0; i < NUM_CTX_PER_RUN / threadcount; i++) {
+        ctx = SSL_CTX_new(TLS_server_method());
```
mattcaswell, line 34: For a server method I think there needs to be at least some minimal configuration for it, e.g. adding a certificate and key. But perhaps we also need to do the same test for the client method - ensuring we configure a trust store etc for a client.

perf/ssl_ctx.c
mattcaswell, line 1: We should add something to the README about this test

mattcaswell (changes requested):

perf/ssl_ctx.c
nhorman: Ok, I can add a command line option to select a client/server tls method, and provide options for a cert/key or keystore

vdukhovni:
One of the tasks here should ideally be to isolate the cost of loading the certificate store in the SSL_CTX, by sharing the store across multiple SSL_CTXs.  The mechanism for this is described in the manpage of SSL_CTX_set1_cert_store(3).  Basically get a handle on the store of the first fully initialised CTX, and then set as the store for all the others.

In other words, compare the cost of creating a fully fleshed SSL_CTX with a private store vs. creating a similar SSL_CTX using a pre-existing store, amortising CAfile loading over multiple SSL_CTX objects. However, there may then be multiple threads accessing that shared store, and that could impact performance of store operations. So a trade-off.

Thus we should also look for any impact on handshake performance when multiple SSL_CTX objects are used, but their underlying cert store is shared.

nhorman:
Is that really what we want here though?  I'm honestly askinig, as I'm not sure.

While I understand that what you're describing is an optimization to creating multiple SSL_CTX objects, this is a performance test.  Do we want to test the optimized path, or do we want to test the worst case path?

I could see an argument for both, but if we're working under the assumption that  the optimized application commonly only creates a single (or very few) SSL_CTX objects that then get shared among multiple SSL objects, I would make an argument that we want to test and optimize the slow path, that rebuilds the store every time, so that the single creation case gets addressed.

Opinions welcome of course.

vdukhovni:
> Is that really what we want here though? I'm honestly askinig, as I'm not sure.
> 
> While I understand that what you're describing is an optimization to creating multiple SSL_CTX objects, this is a performance test. Do we want to test the optimized path, or do we want to test the worst case path?
> 
> I could see an argument for both, but if we're working under the assumption that the optimized application commonly only creates a single (or very few) SSL_CTX objects that then get shared among multiple SSL objects, I would make an argument that we want to test and optimize the slow path, that rebuilds the store every time, so that the single creation case gets addressed.
> 
> Opinions welcome of course.

The point isn't to test the optimisation, rather it is to determine whether, in fact as surmised, the majority of the cost of SSL_CTX creation is loading the CAfile, and therefore any effort to reduce that cost should be spent on decoders, and the store, or whether there are other costs that also matter.

So certainly comparing shared store vs. per-CTX store was/is part of what we want to learn.  Whether this then impacts runtime perf is I guess then interesting, but optional, in that we're now exploring whether to recommend this optimisation, or whether it has a non-trivial downside. Good to learn, but perhaps a separate question?  I'd like to know the answer, but it is a separate question.

nhorman:
But we can't (or well, we can, but arguably shouldn't) create fan out for every combination of operation in one test (that would become unruly very quickly I think).  If you want to know where you spend most of your time in a given set of call paths, use them all and run a tool like perf or gprof to determiine where your time is most frequently spent.

vdukhovni:
> But we can't (or well, we can, but arguably shouldn't) create fan out for every combination of operation in one test (that would become unruly very quickly I think). If you want to know where you spend most of your time in a given set of call paths, use them all and run a tool like perf or gprof to determiine where your time is most frequently spent.

Sure, which is why we can probably skip measuring handshake impact, but if we truly want to understand CTX creation cost, it is I think useful to try to isolate the X509 trust store setup, which is naïvely expected to be the bulk of the cost. Otherwise, what are we really trying to measure here?

nhorman:
My expectation (based on prior comments), was that we were trying to measure the 'cold' creation time of an SSL_CTX.  That is to say, the creation and population of an SSL_CTX with data that would make it usable for handling connections.  The identification of highly latent paths within that context is the (IMHO) the responsibility of adjunct tools like gprof or perf, should we find a performance degradation between compared builds.

vdukhovni:
That's different from the original discussion during the backlog review. I think that isolating the store creation is an obvious high-level knob that is much simpler to turn than gprof, but I'm probably repeating myself. What was the motivation for doing this work? I thought it was reports of regressions in performance of this call, possibly as a result of higher costs in the decoder side, primarily in the CAfile parsing, which splitting it out would expose. Your call...

nhorman:
There were no explicit issues raised regarding the latency of this call (that I am aware of).  Rather the handshake test creates a shared ssl_ctx for all created connections, and the discussion I recall was that:
a) we should have a test which creates an ssl_ctx per connection, rather than a shared one, as doing so is a not uncommon practice in some environments (which we can now toggle with the -s flag in the handshake test)

b) It would be nice to have an ssl_ctx standalone test to measure ssl_ctx_new performance (based on institutional knoweldge that ssl_ctx is generally an expensive operaition).  Thats what this test does.  The comments [here](https://github.com/openssl/tools/pull/196#discussion_r1625762518) suggested that the calling of SSL_CTX_new by itself is in fact not expensive, but the bulk of the latency occurs when adding a keystore/ca/key/etc), and so this test was modified to add those operations.

Its my view that any test should really attempt to test the worst case scenario, from which we can use adjunct tools to identify the slow paths.  Creating efficiencies in our test cases really just puts us in the position of observing that in the nominal/efficient case, performance is ok (unless someone opens an issue that says the nominal/efficient case is somehow bad, in which case, yes, that certainly drives the argument that we should create a test case for that scenario). 

Put another way:  Right now we're working on the assumption that SSL_CTX_new with a per-ctx cert store/cey/cafile/etc is our slow path, and thats what we need to improve.  If there are reports of other scenarios, yes, we should add those test cases as well, but baring that, Adding efficiencies just makes things run faster, and lets us ignore the slow path.

Not saying that we shouldn't have extra tests for various situations, but from my understanding of past discussions adn comments, this is what we want to address right now.

vdukhovni:
Thanks for the clarification, makes sense. Just to make sure we're not miscommunication, I was not suggesting *only* testing the code path where the trust store is shared. Rather, I was anticipating that the test would cover the worst-case scenario you describe, and also, separately (suitable command-line flag) would test with a shared store for comparison to *validate* the intuition that just he store setup accounts for almost all the costs. Perhaps I already made that clear, if so please pardon the repetition.

If you feel that having *both* metrics is not worth the effort, so be it. Thanks for bearing with me so far.

perf/ssl_ctx.c
```diff
@@ -0,0 +1,175 @@
...
+                                              SSL_FILETYPE_PEM) == 1) ||
+                (SSL_CTX_use_PrivateKey_file(ctx, keypath,
+                                             SSL_FILETYPE_PEM) == 1))
+                goto out;
+        } else {
+            if (SSL_CTX_load_verify_dir(ctx, storepath) != 1)
+                goto out;
+        }
+        SSL_CTX_free(ctx);
+        ctx == NULL;
```
Sashan, line 60: perhaps should be `ctx = NULL`. however I think we can just delete the line here.

perf/ssl_ctx.c
```diff
@@ -0,0 +1,175 @@
...
+        usage(argv[0]);
+        goto err;
+    }
+
+    threadcount = atoi(argv[optind]);
+    if (threadcount < 1) {
+        printf("threadcount must be > 0\n");
+        goto err;
+    }
+
```
Sashan, line 129: we should perhaps round-up thread count so we can divide jobs evenly to all cpus. This has been suggested change in this PR: https://github.com/openssl/tools/pull/200#discussion_r1631198883
```
num_ctx_per_run = NUM_CTX_PER_RUN +  (NUM_CTX_PER_RUN % threadcount)
```

Sashan (changes requested):
Looks good in general, have just two minor suggestions.

perf/ssl_ctx.c
```diff
@@ -0,0 +1,175 @@
...
+{
+    OSSL_TIME duration, ttime;
+    uint64_t us;
+    double avcalltime;
+    int ret = EXIT_FAILURE;
+    int i;
+    int terse = 0;
+    int ch = 0;
+
+    while ((ch = getopt(argc, argv, "m:c:k:s:")) != -1) {
```
Sashan, line 91: `tesrs` variable is never set by `getopt()` loop, it remains set to 0. is that intentional?

Sashan (changes requested):

perf/ssl_ctx.c
```diff
@@ -0,0 +1,175 @@
...
+    end = ossl_time_now();
+    times[num] = ossl_time_subtract(end, start);
+
+    if (!ret)
+        err = 1;
+}
+
+static void usage(char *name)
+{
+    printf("usage\n");
```
Sashan, line 73: just a thought: should we print errors and `usage()` output to stderr? The assumption here is that we might building further tooling around those utilities. and it is easier to handle expected output on stdout and error messages in stderr in scripts/tools we might be building on top of those.

vdukhovni: I tend to print *unsolicited* usage messages (invalid command-line options, ...) to `stderr`, while printing *solicited* usage (via `-help` or similar explicit options) to `stdout`, so that one can pipe the latter to `less` (or `more`) without `2>&1` or the like.

perf/ssl_ctx.c
vdukhovni: Sure, only that's not the right expression to round up to the next multiple. Better:

```
/* Assuming `threadcount` is always positive */
int excess = NUM_CTX_PER_RUN % threadcount;
num_ctx_per_run = NUM_CTX_PER_RUN + (excess ? threadcount - excess ? 0);
```

perf/ssl_ctx.c
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    threadcount = atoi(argv[optind]);
+    if (threadcount < 1) {
+        printf("threadcount must be > 0\n");
+        goto err;
+    }
+
+    /* Adjust the number of iterations so we divide evenly among threads */
+    adj_ctx_per_run = (NUM_CTX_PER_RUN/threadcount)*threadcount;
```
mattcaswell, line 134: nit: spaces around `/` and `*`

perf/ssl_ctx.c
```diff
@@ -0,0 +1,181 @@
...
+    start = ossl_time_now();
+
+    for (i = 0; i < adj_ctx_per_run / threadcount; i++) {
+        ctx = SSL_CTX_new(mode == OP_SERVER ? TLS_server_method() :
+                                              TLS_client_method());
+        if (ctx == NULL)
+            goto out;
+        if (mode == OP_SERVER) {
+            if ((SSL_CTX_use_certificate_file(ctx, certpath,
+                                              SSL_FILETYPE_PEM) == 1) ||
```
mattcaswell, line 52: Shouldn't this be `!= 1`? Here and below.

perf/ssl_ctx.c
```diff
@@ -0,0 +1,181 @@
...
+                goto out;
+        }
+        SSL_CTX_free(ctx);
+    }
+
+out:
+    end = ossl_time_now();
+    times[num] = ossl_time_subtract(end, start);
+
+    if (!ret)
```
mattcaswell, line 67: We never assign `ret` to anything and it defaults to 1, so errors will never get reported.

mattcaswell (changes requested):

nhorman:
@mattcaswell given that we're moving these tools to the perf repository, I'm actually going to close this and open a new pr there, with your issues addressed

nhorman:
actually scratch that last comment, i'll fix them up here, it will be much easier for me to port it over once its all fixed up properly

perf/ssl_ctx.c
```diff
@@ -49,23 +48,24 @@ static void do_create_ctx(size_t num)
             goto out;
         if (mode == OP_SERVER) {
             if ((SSL_CTX_use_certificate_file(ctx, certpath,
-                                              SSL_FILETYPE_PEM) == 1) ||
+                                              SSL_FILETYPE_PEM) != 1) ||
                 (SSL_CTX_use_PrivateKey_file(ctx, keypath,
-                                             SSL_FILETYPE_PEM) == 1))
+                                             SSL_FILETYPE_PEM) != 1)) {
+                err = 1;
```
mattcaswell, line 54: There is also a `goto out` in the `if (ctx == NULL)` check above. Shouldn't we set `err = 1` there too?

mattcaswell (changes requested):

mattcaswell (approved):

Sashan (approved):
all looks good, thanks.

t8m:
Hmm.. this PR needs to be transfered over to openssl/perftools repo.

nhorman:
yeah, see my comment [here](https://github.com/openssl/tools/pull/196#issuecomment-2217593861).  My plan is to get these merged, then do a sync pr to perftools, while at the same time doing a PR to remove the code from this one

t8m (approved):

t8m:
IMO there is no point to merge this here. The PR is approved so if you apply identical changes to the perftools repo, you can just merge it there.

vdukhovni:
Is this still pertinent? I can review if that still makes sense.

Sashan:
I think this PR needs to opened against https://github.com/openssl/perftools if we want to move it forward.

t8m:
No longer relevant for this repo.

PR #199, opened 2024-06-06
Fix various issues with the release staging script

t8m:


t8m:
ping @openssl/otc

release-tools/release-aux/release-data-fn.sh
```diff
@@ -8,8 +8,8 @@
 
 # Public or premium release?  Let the version numbers determine it!
 declare -A _ossl_release_types=(
-    [premium]='^1\.0\.2'
-    [public]='^(1\.1\.1|[3-9]\.)'
+    [premium]='^1\.[01]\.[12]'
```
nhorman, line 11: This to me (I think) says premium releases are identified as 1.0.1, 1.0.2, 1.1.1 and 1.1.2

Should that be future proofed for a bit by extending [12] to [123456789] or some such?  Presumably, we're going to bump these patch levels, or are we using alphabetic designation in perpetuity here?

nhorman (commented):
Generally looks good, just the one question

release-tools/release-aux/release-data-fn.sh
t8m: We IMO should not bump the number as that could possibly break things for our customers depending on the fix being marked by the alphabetic suffix. There will be a problem when we reach 1.0.2zzzzzzzzzze release :grin: as that will be the last one that can be represented by the OPENSSL_VERSION_NUMBER. However that will be reached some time near the end of this century if we keep the current rate of patch releases.

nhorman (approved):

release-tools/release-aux/release-data-fn.sh
levitte, line 11: Why not just `'1\.'`?

levitte: I really don't see why we should make it more complicated than necessary...

levitte (approved):

t8m:

PR #200, opened 2024-06-06
Various perf tools improvements

t8m:
- make it easier to build against OpenSSL 1.1.1
- adjust the test iterations to even out the test runtimes
- fix bug in evp_fetch test when running with explicit type


nhorman (approved):

Sashan (approved):

perf/evp_fetch.c
```diff
@@ -223,8 +224,8 @@ int main(int argc, char *argv[])
     }
 
     threadcount = atoi(argv[argnext]);
-    if (threadcount < 1) {
-        printf("threadcount must be > 0\n");
+    if (threadcount < 1 || NUM_CALLS_PER_TEST % threadcount != 0) {
```
Sashan, line 227: I would actually prefer a small tweak here:
```
num_calls_per_test = NUM_CALLS_PER_TEST +  (NUM_CALLS_PER_TEST % threadcount)
```
the idea is to adjust the number of iterations so the job can be always evenly divided. This will allow us to use thread counts 32, 64, and similar. This should implemented for all performance tools.

thanks.

t8m: Makes sense, I'll add it.

t8m:
@nhorman adjusted according to @Sashan suggestion, please reconfirm.

t8m:
@openssl/otc please review.

nhorman:
ACK, approval holds

perf/evp_fetch.c
Sashan: My suggestion is wrong here, my math is weak. Please go for solid suggestion from @vdukhovni mentioned [here](https://github.com/openssl/tools/pull/196#discussion_r1642659934)
```
/* Assuming `threadcount` is always positive */
int excess = NUM_CTX_PER_RUN % threadcount;
num_ctx_per_run = NUM_CTX_PER_RUN + (excess ? threadcount - excess ? 0);
```

levitte (approved):
Sure

Sashan (approved):
rounding up looks good now. thanks.

t8m:

PR #202, opened 2024-06-17
This change brings command line arguments of pkeyread tool one step

Sashan:
closer to other perormance tools. The list of changes in this diff is as follows:
  - stop using -v (a.k.a. verbose) and use -t (--terse) to control output verbosity

  - number of thread to use is mandatory argument now

  - add usage() function which shows hints when invalid arguments are passed.

### DIFF
--- a/perf/pkeyread.c
+++ b/perf/pkeyread.c
@@ -11,6 +11,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


t8m (approved):

nhorman (approved):

mattcaswell (approved):

t8m:
Merged. Thank you.

PR #203, opened 2024-06-17
This change hopes to provide a better alignement of pkeyread with other

Sashan:
performance tools we have. The pkeyread enables as to evaluate speed of reading various keys in der/pem formats. This requires user to specifi desired key and format on command line. This makes pkeyread different to other tools which seem to have a common set of options:
   ./tool_name --terse 2
command line above executes tool tool_name running test using 2 threads with terse output. The idea of this change is to provide implicit mode for pkeyread where key and format is determined from command name. For example to run evaulation for rsa key in pem format using 2 threads one runs tool:
   ./rsa.pem 2
The tool for rsa essentially becames a symlink to pkeyread binary.

This should help us to create us more simple shell scripts to run performacne test. For example to run all tests for pem keys we can do something like this in shell
for perf_tool in *.pem ; do
    ./$perf_tool --terse 2 ;
done

### DIFF
--- a/perf/Makefile
+++ b/perf/Makefile
@@ -1,9 +1,15 @@
-all: randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch
+PKEYREAD_SYMLINKS=pkeyread-dh-pem pkeyread-dhx-pem pkeyread-dsa-pem pkeyread-ec-pem pkeyread-rsa-pem pkeyread-x25519-pem \
+	pkeyread-dh-der pkeyread-dhx-der pkeyread-dsa-der pkeyread-ec-der pkeyread-rsa-der pkeyread-x25519-der
+all: randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch \
+	$(PKEYREAD_SYMLINKS)
+
 # Build target for OpenSSL 1.1.1 builds
-all111: randbytes handshake sslnew newrawkey rsasign x509storeissuer rwlocks pkeyread
+all111: randbytes handshake sslnew newrawkey rsasign x509storeissuer rwlocks pkeyread \
+	$(PKEYREAD_SYMLINKS)
 
 clean:
-	rm -f libperf.a *.o randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch
+	rm -f libperf.a *.o randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch \
+	$(PKEYREAD_SYMLINKS)
 
 CPPFLAGS += -I$(TARGET_OSSL_INCLUDE_PATH) -I.
 # For second include path, i.e. out of tree build of OpenSSL uncomment this:
@@ -12,6 +18,8 @@ CFLAGS += -pthread
 LDFLAGS += -L$(TARGET_OSSL_LIBRARY_PATH) -L.
 # For setting RUNPATH on built executables uncomment this:
 # LDFLAGS += -Wl,-rpath,$(TARGET_OSSL_LIBRARY_PATH)
+
+LN=/bin/ln -s
 
 libperf.a: perflib/*.c perflib/*.h
 	$(CC) $(CPPFLAGS) $(CFLAGS) -c perflib/*.c
@@ -49,3 +57,6 @@ regen_key_samples:
 
 pkeyread: pkeyread.c keys.h libperf.a
 	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -o pkeyread pkeyread.c -lperf -lcrypto
+
+$(PKEYREAD_SYMLINKS): pkeyread
+	$(LN) pkeyread $@
--- a/perf/pkeyread.c
+++ b/perf/pkeyread.c
@@ -11,6 +11,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
+#include <libgen.h>
 #include <openssl/pem.h>
 #include <openssl/evp.h>
 #include <openssl/x509.h>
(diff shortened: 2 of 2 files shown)


Sashan:
rebased with current changes.

perf/Makefile
```diff
@@ -1,9 +1,15 @@
-all: randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch
+all: randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch \
+	dh.pem dhx.pem dsa.pem ec.pem rsa.pem x25519.pem \
+	dh.der dhx.der dsa.der ec.der rsa.der x25519.der
```
t8m, line 3: These are really confusing names for aliases of pkeyread. Would you please consider something like `pkeyread-dh-pem`?

Also on Windows this will not be practical IMO. So perhaps just keeping the key type and formats as arguments and handling this within the perftesting script somehow would make more sense.

t8m (changes requested):

perf/Makefile
Sashan: New version of diff uses suggested naming convention. The whole point of this exercise is to unify command line options handling of performance tools. In my opinion the effort here definitely pays off because  then we can construct performance testing scripts without too much thinking which arguments to pass to each tool. It shortens debugging time spent to get tests running.

My plan for windows is to use `COPY` instead of symlink.

t8m: I am not sure about this approach. IMO it is not scalable. @openssl/otc what are others opinions?

perf/Makefile
```diff
@@ -49,3 +57,39 @@ regen_key_samples:
 
 pkeyread: pkeyread.c keys.h libperf.a
 	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -o pkeyread pkeyread.c -lperf -lcrypto
+
+pkeyread-dh-pem: pkeyread
```
paulidale, line 61: What about converting this lot into:
```make
pkeyread-%: pkeyread
	$(LN) -s pkeyread $@
```

or if the suggestion about using a variable to hold them, use that on the left of the colon.

perf/Makefile
```diff
@@ -1,9 +1,15 @@
...
-all111: randbytes handshake sslnew newrawkey rsasign x509storeissuer rwlocks pkeyread
+all111: randbytes handshake sslnew newrawkey rsasign x509storeissuer rwlocks pkeyread \
+	pkeyread-dh-pem pkeyread-dhx-pem pkeyread-dsa-pem pkeyread-ec-pem pkeyread-rsa-pem pkeyread-x25519-pem \
+	pkeyread-dh-der pkeyread-dhx-der pkeyread-dsa-der pkeyread-ec-der pkeyread-rsa-der pkeyread-x25519-der
 
 clean:
-	rm -f libperf.a *.o randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch
+	rm -f libperf.a *.o randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch \
+	pkeyread-dh-pem pkeyread-dhx-pem pkeyread-dsa-pem pkeyread-ec-pem pkeyread-rsa-pem pkeyread-x25519-pem \
+	pkeyread-dh-der pkeyread-dhx-der pkeyread-dsa-der pkeyread-ec-der pkeyread-rsa-der pkeyread-x25519-der
```
paulidale, line 12: Consider defining a make variable to avoid duplication of all of these names?

perf/pkeyread.c
```diff
@@ -217,6 +218,103 @@ static void usage(char * const argv[])
     } while (*format_name != NULL);
 }
 
+static void implicit_report_result(char * const argv[], int terse)
```
paulidale, line 221: style nit: no space after the `*`

perf/pkeyread.c
```diff
@@ -217,6 +218,103 @@ static void usage(char * const argv[])
...
+{
+    if (terse)
+        printf("%lf\n", get_avcalltime());
+    else
+        printf("%s average time per call: %lfus\n", basename(argv[0]),
+	    get_avcalltime());
+}
+
+static int implicit_main(int argc, char * const argv[])
```
paulidale, line 230: style nit: no space after the `*`

paulidale (changes requested):

perf/Makefile
paulidale: I think it could be made scalable using a make variable to hold the list of possibilities.
Perhaps using some string manipulation functions to get things correct and only have one copy of the list.

Sashan:
new version brings all @paulidale suggestions in.
thanks for the Makefile hint.

perf/Makefile
```diff
@@ -49,3 +57,6 @@ regen_key_samples:
 
 pkeyread: pkeyread.c keys.h libperf.a
 	$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -o pkeyread pkeyread.c -lperf -lcrypto
+
+$(PKEYREAD_SYMLINKS) pkeyread-%: pkeyread
```
paulidale, line 61: Drop the `pkeyread-%` it's not required

perf/Makefile
```diff
@@ -1,9 +1,15 @@
-all: randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch
+PKEYREAD_SYMLINKS=pkeyread-dh-pem pkeyread-dhx-pem pkeyread-dsa-pem pkeyread-ec-pem pkeyread-rsa-pem pkeyread-x25519-pem \
+	pkeyread-dh-der pkeyread-dhx-der pkeyread-dsa-der pkeyread-ec-der pkeyread-rsa-der pkeyread-x25519-der
+all: randbytes handshake sslnew newrawkey rsasign x509storeissuer providerdoall rwlocks pkeyread evp_fetch \
+	$(PKEYREAD_SYMLINKS)
```
paulidale, line 4: Why?

perf/Makefile
```diff
@@ -13,6 +19,8 @@ LDFLAGS += -L$(TARGET_OSSL_LIBRARY_PATH) -L.
 # For setting RUNPATH on built executables uncomment this:
 # LDFLAGS += -Wl,-rpath,$(TARGET_OSSL_LIBRARY_PATH)
 
+LN=/bin/ln
```
paulidale, line 22: This precludes the makefile working on non-unix like targets.  I'm not sure that's a good idea.

perf/Makefile
Sashan: to let make create symlinks as a part of `all` targets.

paulidale: Oops, missed the _all:_

perf/Makefile
Sashan: the easiest hack for platforms with no symlinks we can do is just `LN=copy`.

The current `Makefile` as-is assumes unix-like environment anyway. more work is required to build tools on windows.

given we can not agree on wether implicit arguments approach is acceptable for rwlocks tests here: https://github.com/openssl/tools/pull/205 I'm going to extend run-perf.sh script we discuss here in https://github.com/openssl/tools/pull/207 so it will be able to deal with explicit arguments. I feel this approach to be easier for everyone.

nhorman (approved):

Sashan:
@vdukhovni version of `run-perf.sh` no longer requires this hack to pkeyread. I'd like to close this pull request as unmerged.

PR #204, opened 2024-06-18
All performance tools we have are able to accept --terse

Sashan:
optio. The only exception is handhsake which uses -t. This change makes handshake tool to accept --terse option.

All performance tools report `avcaltime` (avergage call time) when running in --terse mode.  The handsake tool is exception again here. handshake test reports two values avcaltime and handshakes per second. This change makes handshake tool to report avcaltime only when running in --terse mode.

### DIFF
--- a/perf/handshake.c
+++ b/perf/handshake.c
@@ -75,7 +75,7 @@ static void do_handshake(size_t num)
         err = 1;
 }
 
-int main(int argc, char *argv[])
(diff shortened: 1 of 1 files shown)


paulidale (approved):

t8m (approved):

t8m:
Merged with slightly modified commit message. Thank you.

PR #206, opened 2024-06-19
handshake and x509storeissuer tools also require a certsdir argument.

Sashan:
EDIT: Need to think first, then type.

we currently pass threadcount argument as the last argument in all tools. This is inconvenient because it tends to require special handling in scripts built on top of those tools. What I'd like to do is to move _threadcount_ argument which is common to all tools to first position. The _certsdir_ argument is currently used by `handshake` and `x509storeissuer` so it can be moved to second position. This will allow us to execute all tools with common set of options: `toolname --terse 8 /path/to/certsdir`, tool which does not require certsdir can ignore just ignore the next option.

this is simple change which is sufficient. Alternative approach would be to pass path to certificates via environment variable. This would be also fine with me. But I think we can save it for another day.

### DIFF
--- a/perf/evp_fetch.c
+++ b/perf/evp_fetch.c
@@ -10,6 +10,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#include <libgen.h>
+#include <unistd.h>
 #include <openssl/evp.h>
 #include <openssl/kdf.h>
 #include <openssl/core_names.h>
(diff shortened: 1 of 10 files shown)


mattcaswell (approved):

t8m:
Needs rebase.

Sashan:
rebased.

Sashan:
It has turned out more work is required to unify argument handling. This change is mechanical. It introduces `getopt(3)` to all tools to handle command line arguments. Initial version had `getopt_long()` but this is not potable to windows. Therefore --terse option is no longer supported, users need to use short version `-t` instead.

the change is large, but mechanical.

t8m (approved):

mattcaswell (approved):

t8m:
Merged. Thank you.

PR #208, opened 2024-06-25
Add run-perf.sh

vdukhovni:
This is just the run-perf.sh script for sharing with the team, work-in-progress, do not merge.

### DIFF
--- a/perf/run-perf.sh
+++ b/perf/run-perf.sh
@@ -1,0 +1,272 @@
+#!/usr/bin/env bash
+
+#
+#
+# Copyright 2024 The OpenSSL Project Authors. All Rights Reserved.
+#
+# Licensed under the Apache License 2.0 (the "License").  You may not use
+# this file except in compliance with the License.  You can obtain a copy
+# in the file LICENSE in the source distribution or at
+# https://www.openssl.org/source/license.html
+#
+
+#
+# Iterations to run each tool
+#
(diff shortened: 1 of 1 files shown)


Sashan:
I think this is good enough. I gave it try. to speed things up I've modified some parameters:
  - reduced number of iterations to two
  - reduced number of threads to 4 and 8
  - reduced versions to 1.1.1, 3.0, 3.3 and master
 
works flawlessly. I think it's good enough. I would not waste too much time with it. What we really need to decide are parmeters which we will be using for testing on all platforms:
  - number of iterations: the first version from @nhorman  got 50, I've reduced the number to 25, but it still feels too much
  - versions I think @mattcaswell proposal to test 1.1.1, 3.0, 3.3 and master is sufficient
  - number of threads I like what @nhorman suggested: 1 2 4 8 16 32 64 128

with parameters like above we can expect test duration on ryzen 7 with 8 cores roughly 12 hours.

t8m (approved):

nhorman (approved):

Sashan (approved):
it's good enough to go in.

vdukhovni:
Applied a bit more polish and trimmed the version list. Left iterations and threads alone for now, pending future decisions. Should be good to merge.

nhorman:
merged

PR #209, opened 2024-07-05
Add perftools repository to review-tools

nhorman:


review-tools/ghmerge
```diff
@@ -28,9 +29,12 @@ Examples:
   ghmerge --nobuild --target openssl-3.0 --cherry-pick 3 16051 paulidale"
     exit 9
 }
-
 set -o errexit
 
+SELFPATH=$(dirname $(readlink -f $0))
+PERLPATH=$(realpath $SELFPATH/../)
+export PERL5LIB=$PERL5LIB:$PERLPATH:$PERLPATH/OpenSSL-Query/lib/
+
```
t8m, line 37: Should this be in a separate commit/PR?

t8m: I have PERL5LIB=/home/mraz/perl5/lib/perl5 and installed OpenSSL/Query inside there.

t8m: @levitte might have some opinion on this

nhorman: ok, so you independently added the path as part of your bashrc (or whatever shell setup you use), makes sense.

That said, I think, given that the OpenSSL/Query module is housed inside the tools repo, and we have to have that available to run these commands, it would make sense to set the path here, but @levitte you likely know best

levitte: OpenSSL::Query is an installable perl module, just like any other perl module.  The OpenSSL-Query directory is *not* an installation of that module.  Did anyone bother do read OpenSSL-Query/README.md?  It has installation instructions, including how you install it locally for your own user.

levitte: That being said, there are tidbits you can use in perl scripts to affect the module path.  One us the `lib` module (see https://perldoc.perl.org/lib):

``` perl
use lib '.';
```

The other is the FindBin module (see https://perldoc.perl.org/FindBin), which works a bit like using `dirname $0` in a sh script, and more.

We actually use these in certain places, for example OpenSSL's Configure script:

https://github.com/openssl/openssl/blob/c215d75f94fcaa598817e739221f33b71b53fb39/Configure#L15-L16

t8m (approved):

levitte (approved):

t8m:
Merged. Thank you.

PR #211, opened 2024-08-14
Add perf test for EVP_PKEY_derive_set_peer

andrewkdinh:
Adds a performance test profiling the run time of `EVP_PKEY_derive_set_peer`

https://github.com/openssl/project/issues/646


perf/evp_setpeer.c
```diff
@@ -0,0 +1,147 @@
...
+
+#define NUM_CALLS_PER_TEST         1000000
+
+int err = 0;
+
+size_t num_calls;
+static int threadcount;
+static OSSL_TIME *times = NULL;
+
+static unsigned char buf[32] = {
```
Sashan, line 26: What kind of key is kept in `buf` the task mentions to use more key types:
`It should be able to measure the performance on at least DH ffdhe2048, ECDH P-256, P-521, and X25519.`

the code as-is seems to measure only one particular key type.

I would suggest you to take a look at `pkeyread.c` I feel this new test should look quite similar.

Sashan (changes requested):

andrewkdinh:
Completely revamped the `./evp_setpeer` perf test. It is now very similar to how pkeyread perf test is set up and run.

I created a new `perf/genkeys_setpeer.sh` file that can be used to generate a `keys_setpeer.h` file for keys for the test to use. The test will go through the key types (specified as a CLI option) and runs `EVP_PKEY_derive_set_peer` split between however many threads is specified.

```console
$ ./evp_setpeer -k all 2
Average time per dh evp_set_peer call: 4191.861000us
Average time per ec256 evp_set_peer call: 125.292000us
Average time per ec521 evp_set_peer call: 760.357000us
Average time per x25519 evp_set_peer call: 0.957000us
$ ./evp_setpeer -k all -t 2
[dh] 4262.306000us
[ec256] 125.517000us
[ec521] 753.860000us
[x25519] 1.373000us
$ ./evp_setpeer -k x25519 -t 2
[x25519] 1.439600us
```

Sashan:
Changes look god, but....

You need to close this PR and open a new one against https://github.com/openssl/perftools repository.
I'm terribly sorry to miss it earlier.

just fork `...openssl/perftools` copy your changes there and submit new pull request. let's do review there.

sorry for any inconveniences.

PR #214, opened 2024-09-04
Fix the sha sum format to be compatible with coreutils

t8m:
Fixes https://github.com/openssl/openssl/issues/18313


levitte (approved):

t8m:
@levitte please reconfirm, there was a bug in the original solution

levitte (approved):

nhorman (approved):

t8m:

PR #217, opened 2024-09-30
Drop requirement for OTC review

nhorman:
Current bylaws under general policies:
https://github.com/openssl/general-policies/blob/master/policies/committer-policy.md

Indicate the following:
```
All submissions must be reviewed and approved by at least two committers.
Neither of the reviewers can be the author of the submission.
```

Given that, it seems we should adjust our tooling to reflect the lack of need for an OTC approval

### DIFF
--- a/review-tools/gitaddrev
+++ b/review-tools/gitaddrev
@@ -138,13 +138,13 @@ foreach (@ARGV) {
         $WHAT = 'tools';
         $min_authors = 1;
         # openssl/tools is governed by OTC
-        $min_otc = 0;
+        $min_otc = 2;
         $min_omc = 0;
     } elsif (/--fuzz-corpora$/) {
         $WHAT = 'fuzz-corpora';
         $min_authors = 1;
         # openssl/fuzz-corpora is governed by OTC
-        $min_otc = 0;
+        $min_otc = 1;
         $min_omc = 0;
     } elsif (/--perftools$/) {
         $WHAT = 'perftools';
@@ -155,7 +155,7 @@ foreach (@ARGV) {
         $WHAT = 'installer';
         $min_authors = 1;
         # openssl/installer is governed by OTC
-        $min_otc = 0;
+        $min_otc = 1;
         $min_omc = 0;
     } elsif (/^--release$/) {
         $release = 1;
--- a/review-tools/gitaddrev
+++ b/review-tools/gitaddrev
@@ -35,7 +35,7 @@ my $authorcount = 0;
 # be counted as reviewers.  For all other values, it works as a minimum.
 my $min_authors = 0;            # Main source default
 my $otccount = 0;
-my $min_otc = 1;                # Main source default
+my $min_otc = 0;                # Main source default
 my $omccount = 0;
 my $min_omc = 0;                # Main source default
 my $author = $ENV{GIT_AUTHOR_EMAIL};
@@ -138,13 +138,13 @@ foreach (@ARGV) {
         $WHAT = 'tools';
         $min_authors = 1;
         # openssl/tools is governed by OTC
-        $min_otc = 2;
+        $min_otc = 0;
         $min_omc = 0;
     } elsif (/--fuzz-corpora$/) {
         $WHAT = 'fuzz-corpora';
         $min_authors = 1;
         # openssl/fuzz-corpora is governed by OTC
-        $min_otc = 1;
+        $min_otc = 0;
         $min_omc = 0;
     } elsif (/--perftools$/) {
         $WHAT = 'perftools';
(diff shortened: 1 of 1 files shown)


review-tools/gitaddrev
```diff
@@ -138,13 +138,13 @@ foreach (@ARGV) {
         $WHAT = 'tools';
         $min_authors = 1;
         # openssl/tools is governed by OTC
-        $min_otc = 2;
+        $min_otc = 0;
```
mattcaswell, line 141: Is this correct? I'm not necessarily against this change but previously only OTC could commit to this repo. Now any committer can. I'm not sure if that was the intent behind the bylaws change which I think was primarily about the main source repo - and I suppose other repos using the same rules as main source.

nhorman: I think its in keeping with the letter of they bylaws.  As to the spirit, I don't think I can say.  I'd look to other committers/OTC members  on that

t8m: This is the otc/tools repo in the primary GHE server. IMO this is wrong. The bylaws are about the primary openssl/openssl codebase but can be probably extended to all the openssl/<something> repos.

Please revert this hunk.

mattcaswell: > I think its in keeping with the letter of they bylaws

I don't think its in keeping with the letter either. It depends on what you think the scope of that sentence is about. But the policy in general talks about what committers can do, and committers are defined to be "contributors with commit access to the OpenSSL source code repository." - which I think is fairly unambiguously openssl/openssl. Other repos have different review and approval criteria which this policy does not cover.

DDvO:
As mentioned to two project members already in emails on Sep 12 and 13:

With having only one 'review pending' label left on the GitHub GUI,
it is no more possible to quickly see from the labels (e.g., in the GitHub PR list view) whether one or two reviews are pending.
Moreover, 
now it can be rather unclear if a first valid approval is still there, because it may have been invalidated by further activity.
This can lead to confusion and misunderstandings on whether just one or even two approvals are still/again needed.

Therefore, **I propose having two such labels again**, e.g., 'review 1 pending' and 'review 2 pending'.

review-tools/gitaddrev
nhorman: so, whats the right change here?  I can agree that min_otc can/should be 2 for the tools repo, and 0 for openssl, what about perftools/installer/etc?

tom-cosgrove-arm:
+1 to [@DDvO's comment](https://github.com/openssl/tools/pull/217#issuecomment-2383379778) - it's not clear to me now when is best to remove the approval pending label, having two labels was really useful

review-tools/gitaddrev
t8m: I think we can agree that perftools, installer and fuzz-corpora repos should have _similar_ rules as the main source code.

mattcaswell: Strictly speaking the committer policy that was changed *only* applies to openssl/openssl. However, I think the intent for installer was that it should have the same review requirements as openssl/openssl (although I don't know if that was documented anywhere....possibly in some meeting minutes somewhere). I don't know what the right answer for fuzz-corpora is - probably it needs separate discussion.

nhorman: So can we agree here that we should limit this change to the default case, as the other repos are explicitly called out?  Or do we need to bring this up at OTC?

mattcaswell: I think we limit it to just the main repo for now. All other repos require a separate agreement as to what the correct review criteria are.

mattcaswell (approved):

DDvO:
> It is very easy to find out whether additional review is needed or not as GH indicates that via the merge blocked indicator. Or you can just count the green approval check marks.

I suppose you do not mean the "Merging is blocked" but "Review required" indicator, which automatically disappears when all two required reviews are present.
Yet this info cannot be seen from the list of open pull requests, 
unless the "review pending" label gets cleared and "review: done" get added - can this be achieved automatically? 

Moreover, counting the green approval check marks is tedious and can give wrong results due to superseded reviews.
And these are not shown in the list of open pull requests.
So quickly and correctly determining whether one or two reviews are currently needed is even less convenient/straightforward to do.

t8m:
> Moreover, counting the green approval check marks is tedious and can give wrong results due to superseded reviews.

Oh really tedious? I consider it tedious to constantly change the labels for practically no gain.

Why do you need to distinguish PRs where two reviews are missing from PRs when only one review is missing?

When a review is superseded by something different than just  pushing new commits to the PR, it can be dismissed by any committer. However if the PR just needs changes the other option is to clear the review pending label. Of course that requires someone to re-add the label after the required change is pushed.

t8m:
Anyway, this is off topic here.

t8m (approved):
Although I feel we could change also the installer, perftools and fuzz-corpora repos and keep the spirit of the bylaws change, I do not insist.

t8m:
Merged. Thank you.

PR #231, opened 2025-11-14
Add a release validation script

### LINKED ISSUE #1716: Generate and validate an SBOM release artifact during the release process

nhorman:
Currently, part of release process involves (causally) confirming that the release artifacts pass gpg/sha1/sha256sum checks.  It would be nice if we could automate that a bit.  This script allows a user to automatically download draft release artifacts and confirm that their signatures/sha1sum/sha256 sum match those of the included artifact files

It also introduces the optional feature to do a per-file level validation of the release using an SBOM file.  If we choose to include SBOMS in our releases (see
https://github.com/openssl/openssl/pull/29131), this script will for each file in the archive:
1) confirm that a node in the SBOM file exists for it 2) confirm that the sha256sum of the file in the archive matches that of
   the sum included in the SBOM
3) confirm that the sha256sum of the corresponding file from the git
   tree matches that of the sum included in the SBOM

By confirming these three elements we can have a greater degree of confidence that our automated release pipeline did not alter any of our source files from the initial source that we obtained from git.

### DIFF
--- a/release-tools/validate_release.sh
+++ b/release-tools/validate_release.sh
@@ -1,0 +1,172 @@
+#!/bin/sh
+
+####################################################################
+# Usage: validate_release.sh <release-tag>
+#
+# Requirements for usage
+#
+# 1) You need to have the following tools installed
+# a) gpg
+# b) jq
+# c) ghcli
+#
+# Additional requirements
+# 1) you must be authenticated in ghcli as a user that has access to draft releases
+# 2) You must have the openssl public key imported to your gpg key ring
+######################################################################
+
+TEMPDIR=$(mktemp -d /tmp/validation.XXXXXX)
+
+trap "rm -rf $TEMPDIR" EXIT
+
+RELEASE=$1
+
+mkdir $TEMPDIR/assets
+cd $TEMPDIR/assets
+
+# check tool status 
+if ! command -v jq >/dev/null 2>&1; then
+    echo "You must have jq installed to use this tool"
+    exit 1
+fi
+
+if ! command -v gpg >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


release-tools/validate_release.sh
```diff
@@ -0,0 +1,172 @@
...
+
+# Ensure the openssl public key is in our keyring
+gpg --list-keys BA5473A2B0587B07FB27CF2D216094DFD0CB81EF >/dev/null 2>&1
+if [ $? -ne 0 ]; then
+    echo "OpenSSL GPG key not found in keyring, can't validate release"
+    echo "Please import openssls gpg public key from https://keys.openpgp.org/vks/v1/by-fingerprint/BA5473A2B0587B07FB27CF2D216094DFD0CB81EF"
+    exit 1
+fi
+
+# Ensure we are logged in via gh
```
levitte, line 51: Why? There shouldn't be an obligation to be logged in anywhere particular to download a release.

nhorman: because its a requirement of the gh tool to do so.  We could absolutely drop this requirement and download artifacts with curl, it just increases the complexity of the script, as it means we have to compute the url of each artifact and download it individually.

levitte: I'd rather `curl -LO ${URL}`, then:

``` shell
for e in '' .asc .sha1 .sha256; do
    f=$RELEASE.tar.gz$e
    curl -LO https://github.com/openssl/openssl/releases/download/$RELEASE/$f
    if [ $? -ne 0 ]; then
        echo "Could not download $f"
        exit 1
    fi
done
```

PR #233, opened 2025-12-11
remove enable-crypto-mdebug-backtrace from build specification

### LINKED ISSUE #1763: remove crypto-mdebug-backtrace from openssl entirely
We still configure openssl in some CI builds with the crypto-mdebug-backtrace option enabled, but it doesn't do anything (there is no code that gets optionally compiled with it on or off).  We are also close to merging the sslv3 removal feature branch which alters how we handle this option (i.e. its deprecated and changes on that branch disallow its enablement).

Seems like the best thing to do is to remove it entirely, both from openssl and our build tools.

nhorman:
This option hasn't done anything since 1.0.2, and its use is getting removed from openssl, stop using it here too

Fixes openssl/project#1763

### DIFF
--- a/review-tools/opensslbuild
+++ b/review-tools/opensslbuild
@@ -32,8 +32,8 @@ case "$CC" in
 esac
 
 # Set the features we want to enable
-ENABLES="enable-rc5 enable-md2 enable-ec_nistp_64_gcc_128 enable-tls1_3"
-ENABLES="$ENABLES enable-crypto-mdebug enable-crypto-mdebug-backtrace"
+ENABLES="enable-rc5 enable-md2 enable-ec_nistp_64_gcc_128"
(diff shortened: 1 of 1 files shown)


review-tools/opensslbuild
```diff
@@ -33,7 +33,7 @@ esac
 
 # Set the features we want to enable
 ENABLES="enable-rc5 enable-md2 enable-ec_nistp_64_gcc_128 enable-tls1_3"
-ENABLES="$ENABLES enable-crypto-mdebug enable-crypto-mdebug-backtrace"
+ENABLES="$ENABLES enable-crypto-mdebug"
```
t8m, line 36: When you are touching this, can you please also remove the enable-tls1_3 no-op above?

t8m (approved):

nhorman:
merged to master, thank you!

PR #235, opened 2026-02-26
gitaddrev: Update for the REST API changes in cla.int.openssl.org

levitte:
Previously, a github identity could be referred to with a 'github:{id}'
tuple.  In the new CLA app, this is no longer possible, so the command
line syntax starting with '@' is modified to make it work like any
identity given without '@' prefix.

This allows easy pasting of github identities the way any github user
is used to do it.


paulidale (approved):

t8m (approved):

t8m:
Merged. Thank you.

PR #238, opened 2026-03-23
Make ghmerge useful for backports

esyr:
Provide explicit base commit when rebasing the PR on the target branch; also, use the remote ref to avoid potential stray local commits.


paulidale:
What's inadequate with `picktobranch`?

esyr:
Hm?

paulidale:
Apology, it's `pick-to-branch`.  It is designed to do backports.  It's not the same as `ghmerge` but it's easy enough to use.

esyr:
Will take a look at it, thanks!

t8m (approved):
Actually this might be handy sometimes.

review-tools/ghmerge
```diff
@@ -185,11 +185,12 @@ else
     fi
 fi
 
-set -- $(jq -r '[.head.user.login, .head.ref, .head.repo.ssh_url] | join(" ")' $PR_URL_CONTENTS)
+set -- $(jq -r '[.head.user.login, .head.ref, .head.repo.ssh_url, .base.sha] | join(" ")' $PR_URL_CONTENTS)
 
 WHO=$1
 BRANCH=$2
 REPO=$3
+BASE=$4
```
nhorman, line 193: instead of having to manually specify the base commit here for the rebase, would it be worth auto-computing it by running $(git merge-base <base-branch> HEAD), where base-branch is retrieved from the json of the pr being merged?

nhorman (approved):
Just the one question.   I think pick-to-branch is probably sufficient for this work, but this might be a handy additional tool to have in our pocket.

review-tools/ghmerge
t8m: huh, the base branch is retrieved from the github PR here

t8m:
Merged. Thank you.

PR #239, opened 2026-04-09
add workaround for fips-checksums calculating

quarckster:
I tested this change on `openssl-4.0` branch. I unpacked the generated tarball and ran `make diff-fips-checksums`, it finished successfully, `fips-sources.checksums.new`, `fips.checksum.new`, `fips.module.sources.new` are identical to the original files.

### DIFF
--- a/release-tools/stage-release.sh
+++ b/release-tools/stage-release.sh
@@ -567,11 +567,7 @@ $ECHO "== Configuring OpenSSL for update and release.  This may take a bit of ti
 
 ./Configure cc >&42
 
-$VERBOSE "== Checking fips checksums and source file updates"
-
-if grep -q '^update-fips-checksums *:' Makefile; then
-    make update-fips-checksums >&42
-fi
+$VERBOSE "== Checking source file updates and fips checksums"
 
 make update >&42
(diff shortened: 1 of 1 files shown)


release-tools/stage-release.sh
```diff
@@ -576,9 +580,6 @@ make update >&42
 if [ "$next_method" != 'alpha' ] && grep -q '^renumber *:' Makefile; then
     make renumber >&42
 fi
-if grep -q '^update-fips-checksums *:' Makefile; then
-    make update-fips-checksums >&42
-fi
```
t8m, line 581: Thinking about this, moving it might be a little bit fragile. What if we just did `make clean` before `make update-fips-checksums` would it work as well?

t8m (changes requested):

release-tools/stage-release.sh
quarckster: let me check

quarckster: It works too, I pushed a fixup commit.

t8m (approved):

Sashan (approved):
this looks good. thanks.

t8m:
Merged. Thank you.

PR #241, opened 2026-06-08
test-tools: add initial version

bukka:
This adds a new Python test tool with cert setup for indirect crltest.

It allows generating certs introduced in
https://github.com/openssl/openssl/pull/31244


t8m (approved):

test-tools/ossl_test_tools/__main__.py
```diff
@@ -0,0 +1,11 @@
...
+#
+# Licensed under the Apache License 2.0 (the "License").  You may not use
+# this file except in compliance with the License.  You can obtain a copy
+# in the file LICENSE in the source distribution or at
+# https://www.openssl.org/source/license.html
+
+from .cli import main
+
+if __name__ == "__main__":
+    raise SystemExit(main())
```
nhorman: why not just sys.exit(main()) here?

test-tools/uv.lock
```diff
@@ -0,0 +1,174 @@
+version = 1
+revision = 3
+requires-python = ">=3.10"
+
+[[package]]
```
nhorman, line 5: This file seems like it shouldn't be comitted

nhorman (changes requested):

test-tools/uv.lock
bukka: This is usually practice that is recommended pretty much everywhere for application - see for  example https://pydevtools.com/handbook/how-to/how-to-use-a-uv-lockfile-for-reproducible-python-environments/#best-practices-for-reproducible-environments

test-tools/ossl_test_tools/__main__.py
bukka: it's the same thing but can change it for sure.

bukka: Done

test-tools/README.md
```diff
@@ -0,0 +1,48 @@
+# ossl-test-tools
+
+Python tools for OpenSSL's C tests. Edits test artifacts (certs, CRLs,
+keys) into the .c source files in place by variable name.
+
+## Setup
+
+    cd test/pytools
```
Sashan: perhaps this should be
```
cd test-tools/
uv sync
```

Sashan (approved):
looks useful. after adjusting the path in Setup section thing seems to work. thanks.

t8m (approved):

t8m:
As the force push does exactly what @Sashan proposed, assuming his approval.

bukka:
@t8m I guess this can be merged now

t8m:
Merged. Thank you.

PR #244, opened 2026-07-04
test-tools: add statem_clnt_construct_test generator

bukka:
Generate the embedded client certificate and key (kClientCert, kClientKey) used by test/statem_clnt_construct_test.c.

This is for PR https://github.com/openssl/openssl/pull/31861

### DIFF
--- a/test-tools/ossl_test_tools/cli.py
+++ b/test-tools/ossl_test_tools/cli.py
@@ -14,7 +14,7 @@ tool, write the subpackage, import it here, and call its register().
 
 import argparse
 
-from . import crltest, pem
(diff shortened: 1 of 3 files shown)


beldmit:
And here I definitely got feeling that you are reinventing tlsfuzzer...

bukka:
This is just one off cert generator that easily produces certs / keys for embedding in tests so if there is a need to ever re-generate, it's an easy task to do.

nhorman (approved):

t8m (approved):

t8m:
Merged. Thank you.

PR #250, opened 2026-09-15
Update active branches in pick-to-branch script

bernd-edlinger:


review-tools/pick-to-branch
```diff
@@ -45,15 +45,6 @@ esac
...
-    TARGET=openssl-3.0
-    ;;
-*3*2*)
-    TARGET=openssl-3.2
-    ;;
-*3*3*)
-    TARGET=openssl-3.3
-    ;;
 *3*4*)
     TARGET=openssl-3.4
```
paulidale, line 49: This goes EOL next month. Ought it be removed too?

paulidale (approved):

review-tools/pick-to-branch
t8m: There is no big deal if it is removed later. Also 3.6 is going away just a couple of weeks later.

t8m (approved):

t8m:
Merged. Thank you for your contribution.

